PR #2, opened 2021-04-17
Fix  float precision issues

SuperCoder7979:
This PR fixes issues with floats casted as doubles represented as `0.20000000298023224` instead of `0.2F`, for example.


coderbot16 (commented):
Looks good to me, though some tests would probably be nice, if that's feasible

LambdAurora (approved):

KJP12 (approved):

coderbot16 (approved):

PR #4, opened 2021-04-18
Improve output of bitwise operations

SuperCoder7979:
This PR improves two pain points with FernFlower's bitwise operation code, specifically:
* `& 65535` and `& 0xFFFF` being decompiled as `& '\uffff'`
* Hex values used in color bit masking such as `0xFF0000`, `0xFF00`, and `0xFF` not being decompiled properly

These issues have been addressed with this PR.

### DIFF
--- a/src/org/jetbrains/java/decompiler/util/IntHelper.java
+++ b/src/org/jetbrains/java/decompiler/util/IntHelper.java
@@ -29,8 +29,10 @@ public final class IntHelper {
    */
   public static String adjustedIntRepresentation(int value) {
     // Check against the special int values to see if we're one of those, and return
-    if (SPECIAL_VALUES.containsKey(value)) {
-      return SPECIAL_VALUES.get(value);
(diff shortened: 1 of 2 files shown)


src/org/jetbrains/java/decompiler/util/IntHelper.java
```diff
@@ -0,0 +1,39 @@
...
+   * Adjusts the string representation of an int to make it easier to read in certain cases, such as values used
+   * to mask components of an RGB color.
+   * @param value the input number
+   * @return The adjusted string
+   */
+  public static String adjustedIntRepresentation(int value) {
+    // Check against the special int values to see if we're one of those, and return
+    if (SPECIAL_VALUES.containsKey(value)) {
+      return SPECIAL_VALUES.get(value);
+    }
```
kashike, line 34: Just `get` and check for `null` instead - avoids two lookups.

kashike (changes requested):

PR #5, opened 2021-04-19
Fix boolean values assigned to int variables

SuperCoder7979:
This PR fixes a longstanding issue where fernflower would assign boolean values to integer values. It also fixes it in the worst way possible. Sometimes, fernflower would produce a result of `int i = true;` for any nonzero values of i, causing much confusion for new modders. Fixing this will improve the experience of new modders looking at minecraft's code.


LambdAurora (approved):
Hacky but still great :)

TheGlitch76 (approved):
if it works it works

PR #7, opened 2021-04-20
Uninlining: simplify output, add tests and support for more fractional pi multiples

williambl:
This PR replaces the hardcoded `UNIT_CIRCLE` array of valid-for-uninlining fractional multiples of pi with a loop which will find and uninline all multiples with numerator and denominator less than or equal to 12.

It also adds tests for the uninlining of max, min, nan, e, pi, multiples of pi, etc.

It also removes unneeded brackets in the output in some cases, and automatically adds casted double values to float uninlining values.


williambl:
I'm considering changing the max from 12 to 20, for parity with the pi multipliers/divisors above. However, I can't think of a common situation where going above 12 would be needed. Thoughts?

williambl:
Discord consensus was for 12.

SuperCoder7979 (approved):

PR #8, opened 2021-04-21
Apply ForgeFlower's patches

SuperCoder7979:
This PR applies ForgeFlower's patches to our current working tree, to reduce the amount of work needed on our part. Many of their patches have been tried and tested in creating good results for decompiled code, which provides a strong base to work off of. Much more testing and cleanup work needs to be done before it can be merged so the PR is in a draft state at the moment.


src/org/jetbrains/java/decompiler/modules/decompiler/exps/ConstExprent.java
```diff
@@ -353,13 +353,13 @@ else if (UNINLINED_FLOATS.containsKey(floatVal)) {
...
+        return new TextBuffer("0.0F / 0.0");
       }
       else if (floatVal == Float.POSITIVE_INFINITY) {
-        return new TextBuffer("1.0F / 0.0F");
+        return new TextBuffer("1.0F / 0.0");
       }
       else if (floatVal == Float.NEGATIVE_INFINITY) {
-        return new TextBuffer("-1.0F / 0.0F");
+        return new TextBuffer("-1.0F / 0.0");
       }
```
kashike: These changes should be reverted.

https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Float.java#L70-L89

kashike (changes requested):

SuperCoder7979:
Most of the glaring issues have been fixed, and while there are certainly areas for improvement what we have now should be ready to go and can be fixed up in the future, as they do not directly affect the decompilation of the Minecraft codebase. Due to the large scope of this PR I'm not expecting anyone to actually review it, but any and all comments are appreciated.

PR #24, opened 2021-08-20
Dont save directory entries as files

Geolykt:
Given that the affected lines were last changed with https://github.com/QuiltMC/quiltflower/pull/18 and that that PR does not have a description, I have no real idea if I interpreted the use of the method correctly

### DIFF
--- a/src/org/jetbrains/java/decompiler/main/decompiler/ConsoleDecompiler.java
+++ b/src/org/jetbrains/java/decompiler/main/decompiler/ConsoleDecompiler.java
@@ -247,6 +247,9 @@ public class ConsoleDecompiler implements IBytecodeProvider, IResultSaver {
 
   @Override
   public void saveDirEntry(String path, String archiveName, String entryName) {
+    if (entryName.lastIndexOf('/') != entryName.length() - 1) {
+      throw new IllegalArgumentException(entryName + " is not a valid folder name. Folders have to end with a slash otherwise archive managers will treat them as files.");
+    }
     saveClassEntry(path, archiveName, null, entryName, null);
   }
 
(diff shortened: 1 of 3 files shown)


src/org/jetbrains/java/decompiler/main/decompiler/ConsoleDecompiler.java
```diff
@@ -247,6 +247,9 @@ public void createArchive(String path, String archiveName, Manifest manifest) {
 
   @Override
   public void saveDirEntry(String path, String archiveName, String entryName) {
+    if (entryName.lastIndexOf('/') != entryName.length() - 1) {
+      throw new IllegalArgumentException(entryName + " is not a valid folder name. Folders have to end with a slash otherwise archive managers will treat them as files.");
+    }
     saveClassEntry(path, archiveName, null, entryName, null);
```
skyrising, line 253: Wouldn't it be better to just add the `'/'` in the entry name passed to `saveClassEntry`? I'd argue it's `saveDirEntry`s job to make sure it's actually saving it as a directory.

src/org/jetbrains/java/decompiler/struct/StructContext.java
```diff
@@ -165,8 +165,8 @@ public FileVisitResult visitFile(Path path, BasicFileAttributes attrs) throws IO
 
       @Override
       public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) {
-        String dirStr = dir.toString();
-        if (dirStr.length() > 1) unit.addDirEntry(dirStr.substring(1));
+        String dirStr = dir.toString() + "/"; // We are saving folders, not files here Path#toString() treats the path as a file by default.
+        if (dirStr.length() > 2) unit.addDirEntry(dirStr.substring(1));
```
skyrising, line 169: I think that can also be reverted then?

Geolykt: it might be better to keep it just in case another implementation of IResultSaver has a similar issue

skyrising: Then that `IResultSaver` is incorrect.

skyrising: The main problem is that `IResultSaver` has no javadoc (so the method names are the only indication of what they should do) and is generally an awful interface, but we're not gonna fix that today.

skyrising (approved):

SuperCoder7979 (approved):

PR #27, opened 2021-08-31
Add `@Override` support for overrides with tighter return types

williambl:
- Breaks `TestGenericNull`

this PR is nowhere near where it should be, only really made it to show that I'm still working on this and so I can easily compare changes


williambl:
Closing because i've lost interest in continuing this and the project has moved on quite a bit

PR #30, opened 2021-09-05
Add a test case for double casts since they currently decompile to invalid source code

kb-1000:


test/org/jetbrains/java/decompiler/SingleClassesTest.java
```diff
@@ -211,6 +211,7 @@ protected void registerAll() {
     register(JAVA_8, "TestNativeMethods");
     register(JAVA_8, "TestThrowLoop");
     register(JAVA_8, "TestShiftLoop");
+    register(JASM, "TestDoubleCast");
```
skyrising, line 214: Can you add a `// TODO` comment describing the problem here and update the results file so the test passes? That way we keep the tests passing as regression tests and still document the known issues.

skyrising (changes requested):

test/org/jetbrains/java/decompiler/SingleClassesTest.java
kb-1000: Done

skyrising (approved):

SuperCoder7979 (approved):

PR #45, opened 2021-11-01
Refactor: Reduce `BasicBlock` id accesses

Kroppeb:
Removes most direct `id` accesses, unless used to index into a `VBStyleCollection`. 
Usages for dot files are done through a `getDebugId()` method (which just returns the id).


src/org/jetbrains/java/decompiler/code/cfg/BasicBlock.java
```diff
@@ -95,6 +98,10 @@ public void removeSuccessor(BasicBlock block) {
 
   // FIXME: unify block comparisons: id or direct equality
```
skyrising, line 99: I think this change resolves the FIXME?

skyrising (approved):

SuperCoder7979 (approved):

PR #49, opened 2021-11-02
Refactor FinallyProcessor to use LinkedMap/Set

Kroppeb:
Removes the need for sorting by blockid in finally? Seems to not change any test results for me


Kroppeb:
Merged into #45

PR #52, opened 2021-11-13
add pre and post increment and decrement tests

Kroppeb:


testData/src/java8/pkg/TestPPMMBoxed.java
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright 2000-2017 JetBrains s.r.o.
```
skyrising, line 2: Maybe this should be updated to 2021?

skyrising (approved):

SuperCoder7979 (approved):

PR #53, opened 2021-11-13
Replace bytecode mappings with text buffer

Earthcomputer:
This PR replaces BytecodeMappingTracer everywhere with internal tracing inside TextBuffer. This improves maintenance, and also makes way for the line length feature. One important change for developers to take note of, is that after this change, text buffers should be appended to other buffers only by TextBuffer.append(TextBuffer), and NOT by converting to a string and then appending.


Kroppeb (approved):
LGTM

skyrising (approved):

PR #76, opened 2022-01-06
Close opened file systems

CoolMineman:


SuperCoder7979 (approved):

Leo40Git (approved):

src/org/jetbrains/java/decompiler/struct/StructContext.java
```diff
@@ -140,29 +140,20 @@ else if (filename.endsWith(".zip")) {
...
-    FileSystem fs = getZipFileSystem(file);
+    FileSystem fs;
+    try {
+      URI uri = new URI("jar:file", null, file.toURI().getPath(), null);
+      try {
+        fs = FileSystems.newFileSystem(uri, Collections.emptyMap());
+        toClose.add(fs);
+      } catch (FileSystemAlreadyExistsException e) {
+        fs = FileSystems.getFileSystem(uri);
+      }
```
skyrising, line 153: Slight issue: this doesn't work with multiple threads. One could create the new file system, add it to `toClose`; another thread uses the already open file system; the first thread closes it; the second thread tries to use the closed file system.

CoolMineman: Such a project would require a complex synchronization system for FileSystem closing and should just ensure it is open for the duration of decompilation. Similar code is used in stitch and tiny remapper

PR #84, opened 2022-03-14
Adjust IResultSaver to support both Fabricflower and Forgeflower APIs

zml2008:
This should allow VanillaGradle to use Quiltflower, with only minor changes needed on our side.

### DIFF
--- a/src/org/jetbrains/java/decompiler/main/extern/IResultSaver.java
+++ b/src/org/jetbrains/java/decompiler/main/extern/IResultSaver.java
@@ -17,6 +17,9 @@ public interface IResultSaver {
   void copyEntry(String source, String path, String archiveName, String entry);
 
   void saveClassEntry(String path, String archiveName, String qualifiedName, String entryName, String content);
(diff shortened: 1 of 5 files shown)


skyrising (approved):

src/org/jetbrains/java/decompiler/main/extern/IResultSaver.java
```diff
@@ -17,6 +17,9 @@
   void copyEntry(String source, String path, String archiveName, String entry);
 
   void saveClassEntry(String path, String archiveName, String qualifiedName, String entryName, String content);
```
zml2008, line 19:
```suggestion
  void saveClassEntry(String path, String archiveName, String qualifiedName, String entryName, String content);

```

Earthcomputer (approved):

OroArmor (approved):

ByMartrixx (approved):

Geolykt (approved):
It might break ABI, however I am not sure how javac compiles default methods so I am probably worrying about nothing. Either way a step in the right direction

PR #94, opened 2022-04-22
Port dump code lines option from ForgeFlower

jpenilla:
Allows exporting line maps without a custom IResultSaver, which makes it easier to apply the line maps to a jar with a separate application (i.e. [ForgeAutoRenamingTool](https://github.com/MinecraftForge/ForgeAutoRenamingTool/blob/9058118b9182d4a58f128ba3a728e027873742ca/src/main/java/net/minecraftforge/fart/internal/FFLineFixer.java) or [paperweight](https://github.com/PaperMC/paperweight/blob/729ff50a6c7616943bc25e0a08ac6a35549b0f2f/paperweight-lib/src/main/kotlin/tasks/LineMapJar.kt))

original patch: https://github.com/MinecraftForge/ForgeFlower/blob/4fe9bdfec3e6a74e0502afafe0a60f4cdb8a7c85/FernFlower-Patches/0040-Expose-line-mapping-information-in-archive-mode.patch


SuperCoder7979 (approved):
LGTM!

PR #95, opened 2022-04-25
ssa and ssau merge

Kroppeb:
Merges ssa and ssau processing into one class


src/org/jetbrains/java/decompiler/modules/decompiler/vars/VarVersionEdge.java
```diff
@@ -15,6 +15,7 @@ public class VarVersionEdge { // FIXME: can be removed?
   private final int hashCode;
 
   public VarVersionEdge(int type, VarVersionNode source, VarVersionNode dest) {
+    assert source != null && dest != null;
```
SuperCoder7979, line 18: ValidationHelper

src/org/jetbrains/java/decompiler/modules/decompiler/sforms/SSAProcessor.java
```diff
@@ -0,0 +1,964 @@
...
+import org.jetbrains.java.decompiler.modules.decompiler.vars.VarVersionsGraph;
+import org.jetbrains.java.decompiler.struct.StructMethod;
+import org.jetbrains.java.decompiler.struct.gen.MethodDescriptor;
+import org.jetbrains.java.decompiler.util.*;
+
+import java.util.*;
+
+import static org.jetbrains.java.decompiler.modules.decompiler.sforms.VarMapHolder.mergeMaps;
+
+class SSAProcessor {
```
SuperCoder7979, line 18: I think this should be renamed "SFormsConstructor" and made abstract. The reasoning for the name change is that the "Processor" suffix has a specific connotation in the existing QF code and old FF code, so this would add confusion as it's not making structural changes to the statement like the other processors.

src/org/jetbrains/java/decompiler/modules/decompiler/sforms/SSAProcessor.java
```diff
@@ -0,0 +1,964 @@
...
+    }
+  }
+
+  private void processExprent(DirectNode node, Exprent expr, VarMapHolder varmaps, Statement stat, boolean calcLiveVars) {
+
+    if (expr == null) {
+      return;
+    }
+
+    assert varmaps.isNormal();
```
SuperCoder7979, line 207: Merge from #107

src/org/jetbrains/java/decompiler/modules/decompiler/sforms/SSAProcessor.java
```diff
@@ -0,0 +1,964 @@
...
+
+    switch (expr.type) {
+      case Exprent.EXPRENT_IF: {
+        IfExprent ifexpr = (IfExprent) expr;
+        this.processExprent(node, ifexpr.getCondition(), varmaps, stat, calcLiveVars);
+        return;
+      }
+      case Exprent.EXPRENT_ASSIGNMENT: {
+        AssignmentExprent assexpr = (AssignmentExprent) expr;
+        // Kropp : pretty sure ssa(u) gives invalid results if this asserts fails
```
SuperCoder7979, line 217: We found that this was true- I think you should either fix it here or remove the name tag from the start of the comment

SuperCoder7979 (changes requested):
I think the class needs comments merged from #107, but otherwise this is a good refactor

src/org/jetbrains/java/decompiler/modules/decompiler/sforms/SSAProcessor.java
SuperCoder7979: I actually wanted to make another subclass for "virtual" ssa(u) processing that collects information but doesn't actually change the var exprents so that'd be useful there

Kroppeb: useful where?

SuperCoder7979: as in, being able to subclass and such- i personally don't think you should be able to construct a raw SFormsConstructor

SuperCoder7979 (approved):

PR #106, opened 2022-05-03
Fix statement graph validation

SuperCoder7979:
Fixes most issues with statement graph validation, regarding edges and graph structures. Directgraph validation fixes will be in another PR. The ternary processor isn't included here because it's being refactored already in another PR.

### DIFF
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/IfHelper.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/IfHelper.java
@@ -143,8 +143,7 @@ public final class IfHelper {
               ifparent.getFirst().addSuccessor(ifedge);
 
               ifparent.setIfEdge(ifedge);
-            }
-            else {
+            } else {
               ifchild.getFirst().removeSuccessor(ifchild.getIfEdge());
 
               StatEdge ifedge = new StatEdge(StatEdge.TYPE_REGULAR, ifparent.getFirst(), ifinner);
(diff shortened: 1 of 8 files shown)


src/org/jetbrains/java/decompiler/modules/decompiler/IfHelper.java
```diff
@@ -158,7 +157,7 @@ private static boolean collapseIfIf(IfNode rtnode) {
               if (ifinner.hasAnySuccessor()) {
                 StatEdge edge = ifinner.getFirstSuccessor();
                 if (edge.closure == ifchild) {
-                  edge.closure = null;
+                  edge.closure = ifparent;
```
Kroppeb, line 160: Shouldn't we also need to do this with all the edges that have `ifchild` as closure?

src/org/jetbrains/java/decompiler/modules/decompiler/IfHelper.java
```diff
@@ -313,6 +312,7 @@ else if (elsebranch.succs.size() == 1) {
 
           StatEdge newifedge = new StatEdge(StatEdge.TYPE_REGULAR, firstif.getFirst(), second);
           firstif.getFirst().addSuccessor(newifedge);
+          firstif.setIfEdge(newifedge);
```
Kroppeb, line 315: This has also been added on my branch, when we merge, we should probably check that we don't have this being set twice. It wouldn't do any harm but it would look a bit weird :p

PR #114, opened 2022-05-06
feat: SourceFile debug comments

DiamondMiner88:
Unless stripped, the SourceFile attribute contains some pretty valuable information.

Implementation example:
```java
// $FF: compiled from test.java
class Main {
   public static void main(String[] var0) {
      System.out.println("abc");
   }

   // $FF: compiled from test.java
   private class InternalClass {
   }
}
```


SuperCoder7979 (approved):
LGTM

PR #120, opened 2022-05-08
Introduce real lazy loading for the decompile classpath

zml2008:
This reduces memory usage significantly, and may increase performance in cases where the decompiler context is created and destroyed frequently.

I've also bumped the minimum Java version to 11 for easier porting of these changes. To help users who may want to target older Java versions, I've put in support for reading java runtime images from other JDKs besides the one currently running.

There is some minor tweaking of how threading works, to match my open ForgeFlower PR, and remove the need for synchronization on result savers/output sinks.

Going to keep this as draft until things are ready on the ForgeFlower side, so any changes can be replicated in either direction.


skyrising (approved):

src/org/jetbrains/java/decompiler/struct/JarContextSource.java
```diff
@@ -0,0 +1,135 @@
...
+      if (!entry.isDirectory()) {
+        if (name.endsWith(CLASS_SUFFIX)) {
+          classes.add(Entry.parse(name.substring(0, name.length() - CLASS_SUFFIX.length())));
+        }
+        else if (!this.isJar || !name.equalsIgnoreCase(MANIFEST)) {
+          others.add(Entry.parse(name));
+        }
+      }
+      else {
+        directories.add(name);
```
skyrising, line 72: It seems like this doesn't work for all JARs since directory entries are not required to be present

skyrising (changes requested):

src/org/jetbrains/java/decompiler/struct/JarContextSource.java
zml2008: Should be fixed as of 9996e7c

skyrising (approved):

SuperCoder7979 (approved):
Hi, apologies for taking so long to review this PR but it looks all good to me!

SuperCoder7979:
The test fails, but I'll manually fix it after merging.

PR #124, opened 2022-05-09
build: Bump Kotlin version to 1.6.21

zml2008:
This resolves issues with the kotlin compiler daemon on J17 (which should speed up builds a little bit)

As a bonus, the generated bytecode seems to be a fair bit cleaner now


SuperCoder7979 (approved):
Looks good, thank you!

PR #128, opened 2022-05-14
Detect JAR/ZIP files with other file endings

skyrising:
This handles for example self-executing JARs packaged as EXE file,
which contain the JAR content at the end after the native code.

These files are treated like JAR files, meaning any MANIFEST.MF will
not be extracted.


SuperCoder7979 (approved):
LGTM

PR #132, opened 2022-05-18
Replace $FF with $QF

SuperCoder7979:
Purely cosmetic change. Opened as a pull request as it's low priority and I'd like to open it to discussion


l-Luna:
maybe move the "couldn't be decompiled" string to one place so it's not duplicated so much?

skyrising (approved):

PR #135, opened 2022-05-19
Handle JDK constant dynamic bootstraps

l-Luna:
Adds support for `java.lang.invoke.ConstantBootstraps`'s null, enum, primitive class, field, and var handle constant bootstraps.

Introduces a new option/preference/flag (off by default) for decompiling var handle constants.


src/org/jetbrains/java/decompiler/modules/decompiler/CondyHelper.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.Collections;
+import java.util.List;
+
+// Handles the java.lang.invoke.ConstantBootstraps bootstraps
+public class CondyHelper {
+
+  // TODO: handle other bootstraps (invoke, explicitCast)
+  private static final String CONSTANT_BOOTSTRAPS_CLASS = "java/lang/invoke/ConstantBootstraps";
+
+  public static Exprent simplifyCondy(InvocationExprent condyExpr) {
```
SuperCoder7979, line 30: I think this class needs more comments explaining what it's doing, as it's a large amount of code with not much explanation

src/org/jetbrains/java/decompiler/modules/decompiler/CondyHelper.java
```diff
@@ -0,0 +1,180 @@
...
+        if (constArgs.size() != 1) {
+          return condyExpr;
+        }
+        return constructArrayVarHandleExprent(constArgs.get(0));
+      }
+    }
+    return condyExpr;
+  }
+
+  private static Exprent constructVarHandle(String fieldName, String fieldOwner, PooledConstant fieldType, boolean isStatic) {
```
SuperCoder7979, line 101: Same with these methods down here

src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
```diff
@@ -242,6 +242,10 @@ public interface IFernflowerPreferences {
   @Description("Add debug comments showing the class SourceFile attribute if present.")
   String SOURCE_FILE_COMMENTS = "sfc";
 
+  @Name("Decompile complex constant-dynamic expressions")
+  @Description("Some constant-dynamic expressions can't be converted to a single Java expression with identical run-time behaviour. This decompiles them to a similar non-lazy expression, marked with a comment.")
+  String DECOMPILE_COMPLEX_CONDYS = "dcc";
```
SuperCoder7979, line 247: Should be added to readme as well

SuperCoder7979 (changes requested):
Looks good! Only a few minor tweaks with documentation are needed :)

SuperCoder7979 (approved):
Looks good. Thanks for the PR!

PR #139, opened 2022-05-20
Try to fix trycatch blocks that have the same exception type defined multiple times

jaskarth:
This is a last-ditch recovery attempt, essentially. It's not ideal that this has to be a step in the processing pipeline, but I wasn't able to find a general solution for it on the control flow graph stage.

### DIFF
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/TryHelper.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/TryHelper.java
@@ -8,10 +8,7 @@ import org.jetbrains.java.decompiler.struct.StructClass;
 import org.jetbrains.java.decompiler.util.DotExporter;
 import org.jetbrains.java.decompiler.util.StartEndPair;
 
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 
 public class TryHelper {
   public static boolean enhanceTryStats(RootStatement root, StructClass cl) {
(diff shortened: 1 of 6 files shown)


src/org/jetbrains/java/decompiler/modules/decompiler/TryHelper.java
```diff
@@ -185,4 +182,69 @@ private static boolean collapseTry(CatchStatement catchStat) {
...
+
+      Set<List<String>> setExec = new HashSet<>(catchStat.getExctStrings());
+
+      // Check if list of catch headers has a duplicate type
+      if (setExec.size() < catchStat.getExctStrings().size()) {
+        // To not have to calculate catch dominators, split along every catch type for now
+        // TODO: better impl that only splits the needed type
+
+        // Stack of nested statements
+        List<Statement> catchStack = new ArrayList<>();
```
skyrising, line 226: Looks like you only use the last element/top of the stack, so this could probably be replaced with just `Statement`

Kroppeb:
What would this do on the following (jasm) code
```java
try {
  ...
} catch (Exception e) {
  ...
  throw e;
} catch (Exception e) {
  return null;
}
```
cause that would behave differently than
```java
try {
  try {
    ...
  } catch (Exception e) {
    ...
    throw e;
  }
} catch (Exception e) {
  return null;
}
```

SuperCoder7979:
I'm pretty sure it'd split them. How do you think we should handle behavior breaking code like that? Could we search for throws in the catch statement?

Kroppeb:
The problem is that `throw` isn't the only thing that throws, any instance field or instance call can easily throw an npe, and methods can't throw anything. The problem when a later catch exception is dominated by the earlier catches, than it's essentially dead code.

SuperCoder7979:
yeah, that's fair. we could put this behind a preference that is default on perhaps, as it's wanted in some cased and not in others

jaskarth:
This branch has diverged significantly and I think we need a better solution here, so I'll close this.

PR #145, opened 2022-05-24
Add support for static and transitive modifiers of 'requires' in module-info

### LINKED ISSUE #144: `requires static` is decompiled as `requires` in module-info
Example: `META-INF/versions/16/module-info.class` in https://github.com/Juuxel/unprotect/releases/download/1.1.0/unprotect-1.1.0.jar

The static vs normal `requires` info is available in the bytecode since JDK's javap can read it. (And produce a more accurate decompiled version of the module-info, ironically)

kb-1000:
Fixes #144


SuperCoder7979 (approved):
Looks good, thanks!

PR #148, opened 2022-05-25
Don't run processing steps on statements the method doesn't have

SuperCoder7979:
Probably doesn't impact performance greatly, but may reduce iteration allocations for `new ArrayList<>(st.getStats())`.

### DIFF
--- a/src/org/jetbrains/java/decompiler/main/rels/MethodProcessorRunnable.java
+++ b/src/org/jetbrains/java/decompiler/main/rels/MethodProcessorRunnable.java
@@ -167,6 +167,8 @@ public class MethodProcessorRunnable implements Runnable {
       decompileRecord.add("RemoveSynchronizedHandler", root);
     }
 
+    root.buildContentFlags();
+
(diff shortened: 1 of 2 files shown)


skyrising (approved):

src/org/jetbrains/java/decompiler/main/rels/MethodProcessorRunnable.java
```diff
@@ -219,6 +221,10 @@ public static RootStatement codeToJava(StructClass cl, StructMethod mt, MethodDe
 
       // Merge loop
       while (true) {
+        if (!root.hasLoops()) {
```
Kroppeb, line 224: why break out? wouldn't it be better to put the if around the `while (true)`?

SuperCoder7979: Hm. Good point. I thought it'd be cleaner, but I forgot about the other thing at the bottom

Kroppeb (approved):

SuperCoder7979:
do I get extra points if I say "Looks good, thanks!" to myself

PR #149, opened 2022-05-25
Guarded switch cases

l-Luna:
Adds handling for guarded switch cases (`case a && b`, or `case a when b`).

Also fixes type switches in loops without guards.


SuperCoder7979:
As discussed on discord, this PR is blocked by a lack of case exprent processing in SFormsConstructor, which itself is blocked by a lack of representation of case exprents in DirectGraph.

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchPatternMatchProcessor.java
```diff
@@ -48,36 +49,118 @@ private static boolean processStatement(SwitchStatement stat, Statement root) {
...
+              && assignStat.getExprents().size() == 1
+              && assignStat.getParent() instanceof IfStatement
+              && assignStat.getParent().getParent() instanceof SequenceStatement
+              && assignStat.getParent().getParent().getParent() == stat) {
+            Statement next = assignStat.getSuccessorEdges(StatEdge.TYPE_CONTINUE).get(0).getDestination();
+            if (next == stat.getParent()) {
+              IfStatement guardIf = (IfStatement) assignStat.getParent();
+              if (guardIf.getHeadexprent().getCondition() instanceof FunctionExprent) {
+                FunctionExprent cond = (FunctionExprent) guardIf.getHeadexprent().getCondition();
+                Exprent guardExprent = cond.getLstOperands().get(0);
```
l-Luna, line 104: note to self: this is just wrong

l-Luna:
merging with master broke stuff, don't merge yet

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchPatternMatchProcessor.java
```diff
@@ -48,60 +47,173 @@ private static boolean processStatement(SwitchStatement stat, Statement root) {
...
       }
 
-      // Make instanceof
+      // find the pattern variable assignment
+      if (caseStat instanceof SequenceStatement) {
+        Statement oldStat = caseStat;
+        caseStat = caseStat.getStats().get(0);
+        // we can end up with a SequenceStatement with 1 statement from guard `if` elimination, eliminate the sequence entirely
+        if (oldStat.getStats().size() == 1) {
+          oldStat.replaceWith(caseStat);
```
SuperCoder7979, line 185: If oldStat isn't used here afterwards, I think it's better to let SequenceHelper deal with it instead- it should be called after switch pattern processing is done

Mainly because that is it's entire goal, to clean up weird stuff with sequences afterwards so that the actual pipeline passes don't need to deal with it (and because it does some other stuff, like eliminating empty basic blocks that can be inlined)

SuperCoder7979 (changes requested):
Hey, sorry for the delayed review! I have a few questions and a few things for you to look at, mainly some more nuanced details with the statement graph :)

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchPatternMatchProcessor.java
```diff
@@ -144,21 +239,14 @@ private static boolean processStatement(SwitchStatement stat, Statement root) {
...
-        Pair<Statement, Exprent> ref = references.get(0);
-        if (ref.b instanceof AssignmentExprent && ((AssignmentExprent) ref.b).getLeft().equals(var)) {
-          ref.a.getExprents().remove(ref.b);
-        }
+
+    if (guarded && stat.getParent() instanceof DoStatement) {
+      // remove the enclosing while(true) loop of a guarded switch
+      stat.getParent().replaceWith(stat);
+      for (StatEdge edge : stat.getPredecessorEdges(StatEdge.TYPE_CONTINUE)) {
+        stat.removePredecessor(edge);
```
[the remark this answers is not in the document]
Kroppeb: I'd prefer `edge.remove()`

Kroppeb (changes requested):
It seems you are removing stuff before checking that all assignments to `idx` can be removed. There also seem to be some other checks missing that I think should be there, like fall through, and whether the assignment of idx is on it's own, followed by a continue to the loop. I think with some jasm, you could produce a bunch of cases that can trigger this transformation, resulting in either illegal code, or code that is not equivalent, but I think even from java I can make some more failing test cases. (I'll spend some time on a few)

SuperCoder7979 (approved):
Approving this as it's good enough to start testing with, and it's for a preview feature anyway. Thank you so much for spending time on this, it's a tremendous help!

PR #152, opened 2022-05-26
Prevent merging of clashing types, when they are not correctly assignable to the other

SuperCoder7979:
This primarily prevents the merging of int and boolean types, which tended to get merged together even when it is not possible to assign them. This also creates a regression in `TestInfiniteLoop` where a nonone constant becomes boolean `true`, when it shouldn't be. This is likely due to the variable not having any uses. That bug will be addressed in an upcoming PR.


Kroppeb (approved):
lgtm

PR #154, opened 2022-05-26
Fix loops from extending too far due to postdominance not discovering all possible nodes

SuperCoder7979:
This PR fixes the longstanding bug where too many loops would get created due to an issue with postdominance not discovering all possible nodes to extend a new general statement over. It also adds new tracing over DomHelper to better trace it's logic and find problems. This PR also eliminates entirely the "baby shark" issue.

Future work before this PR can be undrafted:
- Find out why adding the loop body to the header's postdom set causes strange issues
- Add comments explaining what everything is doing
- Test on nested loops with switches. There are some weird regressions there (not in the test suite yet)
- Remove debug code left in the classes


Leo40Git (approved):

SuperCoder7979:
Opening this up for review to see if it can get merged relatively soon. There are a few problem cases with nested loops, but it's not a major issue and can be fixed later.

Kroppeb (approved):

PR #159, opened 2022-05-28
Don't process direct nodes when we've seen them already from another node in ExprProcessor

SuperCoder7979:
This fixes an OOM that we have been seeing, where calling this from within FinallyProcessor could lead to many new objects being created, overloading the garbage collector. I'm not 100% confident that this is the right fix, so I'll test a bit more on my own.


Kroppeb (approved):
LGTM

PR #169, opened 2022-06-20
Add test for inverted float comparisons, where early exit would be cleaner

Earthcomputer:


Earthcomputer:
(Motivation: I came across this case several times in the past, and today came across it while cleaning up the GeodeFeature source)

SuperCoder7979:
May be useful to have a default-on preference for this as well

SuperCoder7979 (approved):

PR #179, opened 2022-07-22
Added Implementation-Version to manifest

nbauma109:
1) To make the version number available from inside running application. (useful for an about dialog in a GUI displaying the version number)

2) If the jar was renamed for whatever reason, the version can still be found in the manifest,


Geolykt (approved):
Makes sense to me

SuperCoder7979:
Merging into 1.9.0, thanks for this PR!

PR #184, opened 2022-07-31
Use SSAU to validate finally blocks with variables inside

SuperCoder7979:
This PR uses SSAU to resolve differences in variable indices between finally duplicated blocks.
Also marks semaphore variables more clearly in the code.


SuperCoder7979:
This PR should be squashed

src/org/jetbrains/java/decompiler/modules/decompiler/sforms/SimpleSSAReassign.java
```diff
@@ -0,0 +1,105 @@
...
+      int index = var.getIndex();
+
+      // Skip stack vars
+      if (index < VarExprent.STACK_BASE) {
+        vers.add(index);
+      }
+    });
+
+    // Make new variables to increment the index for
+    int maxVer = vers.stream().max(Integer::compareTo).orElse(-1);
```
Kroppeb, line 32: Why not use the varcounter?

Kroppeb (approved):

PR #187, opened 2022-08-01
Add Scala tests

l-Luna:
Adds tests for code written in Scala.

These use a slightly modified `dec` format that includes the decompiled output of all dependency classes, with a comment specifying their original file; Scala doesn't (seem to?) mark them as inner classes, and they contain most of the logic.

After downloading the compiler and library, extra build time is negligible; YMMV.


SuperCoder7979 (approved):

SuperCoder7979:
Seems like a few tests are failing, likely from my merge. I'll squash and fix them in-dev.

PR #188, opened 2022-08-07
Improve noargs message and add a help argument + detailed message

sschr15:


src/org/jetbrains/java/decompiler/main/decompiler/ConsoleDecompiler.java
```diff
@@ -60,10 +62,119 @@ else if (args.length > x+1) {
     }
     args = params.toArray(new String[params.size()]);
 
+    if (Arrays.stream(args).anyMatch(arg -> arg.equals("-h") || arg.equals("--help") || arg.equals("-help"))) {
+      String[] help = {
```
SuperCoder7979, line 66: This array should be private static final, or the like

src/org/jetbrains/java/decompiler/main/decompiler/ConsoleDecompiler.java
```diff
@@ -60,10 +62,119 @@ else if (args.length > x+1) {
     }
     args = params.toArray(new String[params.size()]);
 
+    if (Arrays.stream(args).anyMatch(arg -> arg.equals("-h") || arg.equals("--help") || arg.equals("-help"))) {
```
SuperCoder7979, line 65: The content of this if body should be moved out to another method (or maybe even another class) because the main method is getting very long

SuperCoder7979 (changes requested):
Thank you so much! I have a few comments to make.

SuperCoder7979 (approved):
Looks good, thanks!

PR #195, opened 2022-08-23
fix: rename entities obfuscated class detection

DiamondMiner88:
Previously, it would not detect classes like `a$a` or `aaa$aaa` which caused some issues.
I am not sure if a cache would be needed since I am using regex now.
As an additional thought, the regex could be a cli paramater since names could be obfuscated differently.


SuperCoder7979 (approved):
Looks good! Is it possible to attach a unit test as well?

DiamondMiner88:
I don't know how unit tests work in quilt sooo

SuperCoder7979:
Fair, I realized it's never specified anywhere. I'll add one myself.

Issue #112, opened 2022-05-05
Rename entities feature causes many problems and doesn't work correctly


SuperCoder7979:
Rename entities is an old feature with many bugs. As there are a lot of bugs, this issue will track them:
- [ ] Fields and method definitions aren't renamed properly
- [ ] Inner classes aren't run with the renamer
- [ ] Renamer too strict on what classes should be renamed
- [ ] Renaming certain classes fails to output them

DiamondMiner88:
Additionally documenting here; inner classes of renamed parents are not inlined with rename entities enabled and are instead written to a separate file. However, references to them seem to be ok so it's not too high priority.

PR #196, opened 2022-08-23
fix: rename entities invalid cache key

DiamondMiner88:
It seemed like an NPE came from StructMethods of renamed classes, the renamed class of which doesn't exist in the StructContext class cache.

Fixes point 4 of #112

### DIFF
--- a/src/org/jetbrains/java/decompiler/struct/gen/MethodDescriptor.java
+++ b/src/org/jetbrains/java/decompiler/struct/gen/MethodDescriptor.java
@@ -90,8 +90,15 @@ public final class MethodDescriptor {
         }
         else if (init) {
           // && isEnum
-          if (DecompilerContext.getOption(IFernflowerPreferences.DECOMPILE_ENUM) && DecompilerContext.getStructContext().getClass(struct.getClassQualifiedName()).hasModifier(CodeConstants.ACC_ENUM)) {
-            actualParams -= 2;
+          if (DecompilerContext.getOption(IFernflowerPreferences.DECOMPILE_ENUM)) {
+            String className = struct.getClassQualifiedName();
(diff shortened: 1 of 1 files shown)


src/org/jetbrains/java/decompiler/struct/gen/MethodDescriptor.java
```diff
@@ -90,8 +90,15 @@ public static MethodDescriptor parseDescriptor(StructMethod struct, ClassNode no
         }
         else if (init) {
           // && isEnum
-          if (DecompilerContext.getOption(IFernflowerPreferences.DECOMPILE_ENUM) && DecompilerContext.getStructContext().getClass(struct.getClassQualifiedName()).hasModifier(CodeConstants.ACC_ENUM)) {
-            actualParams -= 2;
+          if (DecompilerContext.getOption(IFernflowerPreferences.DECOMPILE_ENUM)) {
+            String className = struct.getClassQualifiedName();
+            if (DecompilerContext.getOption(IFernflowerPreferences.RENAME_ENTITIES) && (className = DecompilerContext.getPoolInterceptor().getOldName(className)) == null) {
```
SuperCoder7979, line 95: The inline assignment to `className`  should be replaced, as I think it hurts code readability. It'd be better to use a compound if statement here.

SuperCoder7979 (changes requested):
Looks good, I only have one minor change.

src/org/jetbrains/java/decompiler/struct/gen/MethodDescriptor.java
DiamondMiner88: I assume my new commit is what you mean

SuperCoder7979 (approved):

PR #197, opened 2022-08-27
fix: Correct improper class paths from dir source

zml2008:
old:

```
INFO:  Loading Class: /FuzzedClass_21
```

new:
```
INFO:  Loading Class: FuzzedClass_21
```

This only impacts items sourced from directories, not jars.


SuperCoder7979:
LGTM!

PR #199, opened 2022-09-27
Record pattern tests

l-Luna:
Adds java 19 preview tests with record patterns. Most of these decompile fairly straightforwardly.

The decompiled code is invalid as the proxy methods aren't shown, but still called. `TestRecordPattern4.test2` is also wrong at the end, reusing `var21` where it should not (should be the value of `c`, runs correctly).


l-Luna:
also `TestRenameEntities` magically breaks in CI but not in dev, somehow

SuperCoder7979:
Looks good to me- `TestRenameEntities` is very broken and can be fixed in-tree later

SuperCoder7979:
Can you also add comments to SingleClassesTest describing issues with specific tests and such, if any exist past "record patterns don't work"?

PR #200, opened 2022-10-05
Add test for array constructor references

Juuxel:
See discord discussions. I think it would be good to resugar it to `int[]::new` instead of a lambda.


Earthcomputer:
Maybe add `int[][]::new` to this list?

Juuxel:
Sure. My idea was "primitive array" + "object array" + "array array", but a separate case for $\textsf{primitive array}^2$ might be good to have too.

SuperCoder7979:
Thanks for the extra tests!

PR #215, opened 2022-11-09
Update issue templates

jaskarth:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,17 @@
...
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Quiltflower version**
+Replace this text with the version of Quiltflower that you are using.
+
+**Describe the bug**
```
ByMartrixx, line 13: I think the bold text should be a header instead

jaskarth: I agree. I'll make this change.

PR #233, opened 2022-12-20
add "warn about inconsistent generic signatures" option

ix0rai:
~~This is required for QM to generate MC source without warnings cluttering the logs.~~ Fixed in a different area! I tested and seems to work fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ The rest of options can be left as they are: they are aimed at professional reve
 - thr: maximum number of threads (default is number of threads available to the JVM)
 - jvn (0): Use jad variable naming for local variables
 - sef (0): Skip copying non-class files from the input folder or file to the output
+- wgs (1): Warn about inconsistent generic signatures
 - win (1): Warn about inconsistent inner class attributes
 - dbe (1): Dump bytecode on errors
 - dee (1): Dump exceptions on errors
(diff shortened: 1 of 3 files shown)


OroArmor (approved):

zml2008:
I'm curious - what in QM's process produces these inconsistent signature warnings? I'm pretty sure we don't see them in a decompile with mojang names.

ix0rai:
they occur in the `eraseBytecode` and `genFakeSource` tasks while building a QM javadoc jar, as you can see in [this log](https://github.com/QuiltMC/quilt-mappings/actions/runs/3735940678/jobs/6339735423)

ByMartrixx:
For a bit more context, the class files are processed by [a class adapter that erases most of their code](https://github.com/QuiltMC/javadoc-draftsman/blob/d975b0085c6edafd3086140a4cdd2843a3929606/src/main/java/org/quiltmc/draftsman/asm/DraftsmanClassAdapter.java), which allows us to generate javadocs from a decompilation containing only the relevant information

zml2008:
Thanks for that link - I'm just poking through on mobile, but my suspicion is that when you strip the thrown exception types from constructors, you don't also transform the generic signature to remove those thrown types. If that is true, imo a better solution would be to run those signatures through a stripping process (asm has Signature{Reader,Writer,Visitor} to help with that) to align them with the modified declaration.

ByMartrixx:
The issue was fixed in draftsman with commit QuiltMC/javadoc-draftsman@a66b1de9e944372de701e35d4695ba55f819ea36

ix0rai:
^ this is no longer required by QM, it can be left up to the Quiltflower team on whether or not this is something you want

jaskarth:
I the interest of not creating more options to maintain, I'll close this. Thank you for looking into the issue though!

PR #234, opened 2022-12-23
Move null checks to two equals

sschr15:
Also updates one test that was failing before I changed anything


sschr15:
hec that's wrong

PR #235, opened 2022-12-28
Add long option names & update help message

sschr15:
This also may allow for easier work on combination flags, like perhaps a `--no-removal` which would apply `--no-remove-bridge --no-remove-synthetic --no-remove-getclass --no-remove-empty-try-catch --no-remove-imports` or something like that

### DIFF
--- a/src/org/jetbrains/java/decompiler/main/decompiler/OptionParser.java
+++ b/src/org/jetbrains/java/decompiler/main/decompiler/OptionParser.java
@@ -1,0 +1,79 @@
+package org.jetbrains.java.decompiler.main.decompiler;
+
+import org.jetbrains.java.decompiler.main.extern.IFernflowerPreferences;
+
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+
+public class OptionParser {
+  private static final Map<String, String> SHORT_TO_LONG_NAME_MAP = new HashMap<>();
+
+  private static final String[] CUSTOM_CHECKS = {
+    "-e=",
+    "--add-external=",
+    "-only=",
+    "--only=",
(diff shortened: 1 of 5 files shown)


src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
```diff
@@ -14,207 +14,258 @@
 public interface IFernflowerPreferences {
   @Name("Remove Bridge Methods")
```
ByMartrixx, line 15: The changes made here break any application that passes options to the BaseDecompiler constructor, like the Intellij plugin or Enigma

src/org/jetbrains/java/decompiler/main/decompiler/OptionParser.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Map;
+
+public class OptionParser {
+  private static final Map<String, String> SHORT_TO_LONG_NAME_MAP = new HashMap<>();
+
+  private static final String[] CUSTOM_CHECKS = {
+    "-e=",
+    "--add-external=",
+    "-only=",
+    "--only=",
```
ByMartrixx, line 16: `--only=` isn't handled in `ConsoleDecompiler`

ByMartrixx (changes requested):

ByMartrixx (approved):
Looks good!

Earthcomputer (approved):
These changes are also good for the intellij plugin. I didn't like having to guess types

jaskarth (approved):
LGTM, tysm for doing this! The README needs to be updated, but that can be lower priority.

PR #82, opened 2022-03-03
Refactor/move if pattern matching earlier

Kroppeb:
90% of the lines changed are tests.

PR #254, opened 2023-04-10
Fix some issues with simple patterns

Kroppeb:
Replaces #82

Using this system should make it more friendly to also handle record patters or whatever they are gonna throw at us next, which might need multiple passes which wouldn't have worked with #82 I think


Kroppeb:
ugh, validation isn't passing

src/org/jetbrains/java/decompiler/modules/decompiler/IfPatternMatchProcessor.java
```diff
@@ -47,86 +54,121 @@ private static boolean matchInstanceofRec(Statement statement, RootStatement roo
...
+
+    Exprent source = iof.getLstOperands().get(0);
+    Exprent target = iof.getLstOperands().get(1);
+
+    // Check to make sure there is more than 1 exprent.
+    // More often than not, when there's less than 1 it means it's assigning into a previous value.
+    // TODO: this isn't always the case, handle it properly
+    Statement head = branch.getBasichead();
+
+    if (head == null || head.getExprents() == null) {
```
jaskarth, line 111: `basichead` for an if statement should never be false, so the first branch isn't needed. We should add an assert to getBasicHead validating this invariant.

src/org/jetbrains/java/decompiler/modules/decompiler/stats/IfStatement.java
```diff
@@ -489,4 +489,44 @@ public boolean match(MatchNode matchNode, MatchEngine engine) {
     Integer type = (Integer) matchNode.getRuleValue(MatchProperties.STATEMENT_IFTYPE);
     return type == null || this.iftype == type;
   }
+
+  public boolean checkInversion() {
```
jaskarth, line 493: I think the name needs to be more descriptive, as `check` implies to me that it's pure. Perhaps something like `tryInvert`?

testData/results/pkg/TestPatternMatching17FakeNew.dec
```diff
@@ -15,58 +15,37 @@ public class TestPatternMatching17FakeNew {
...
    }// 32
 
    void test4(Object o) {
-      if (o.hashCode() >= 0) {// 35
-         if (!(o instanceof String)) {
-            return;// 39
-         }
-
-         String var2 = (String)o;
+      if (o.hashCode() < 0 || o instanceof String var2) {// 35
```
jaskarth, line 34: Regression?

jaskarth (changes requested):
Thanks again for taking this on! The lack of reasoning around boolean types was one of the deficiencies with the original system and I'm glad this fixes most of that. I have some comments on the code.

testData/results/pkg/TestPatternMatching17FakeNew.dec
Kroppeb: Nope, that's what the source code. I'm not sure which part you think here regressed?

jaskarth: Oh oops, I didn't have the code in front of me and presumed it was a regression. Nevermind then!

src/org/jetbrains/java/decompiler/modules/decompiler/stats/IfStatement.java
Kroppeb: Mmh, It only inverts when it "needs" to. `tryInvert` sounds like it will invert if possible, even when it isn't needed.

Kroppeb: what about something like `fixInversion`

src/org/jetbrains/java/decompiler/modules/decompiler/IfPatternMatchProcessor.java
```diff
@@ -23,7 +25,13 @@ public static boolean matchInstanceof(RootStatement root) {
     boolean res = matchInstanceofRec(root, root);
 
     if (res) {
-      SequenceHelper.condenseSequences(root);
+      ValidationHelper.validateStatement(root);
+      //noinspection StatementWithEmptyBody
+      if (IfHelper.mergeAllIfs(root)) {
```
jaskarth, line 30: This if statement should be inverted

src/org/jetbrains/java/decompiler/modules/decompiler/IfPatternMatchProcessor.java
```diff
@@ -149,4 +197,91 @@ private static void findVarsInPredecessors(List<VarVersionPair> vvs, Statement r
...
+            break;
+          }
+          case BOOL_NOT: {
+            // when `!` returns true, the exprent had to return false
+            // when `!` returns false, the exprent had to return true
+            return getLastExprentWhen(func.getLstOperands().get(0), !ifTrue, onlyIfTrue);
+          }
+
+          // TEMPORARY
+          // (Nothing is as permanent as a temporary solution :p)
```
jaskarth, line 239: Clean up comment here

jaskarth (changes requested):
Two minor nitpicks this time! I'll run it against Minecraft and see what the diff looks like.

src/org/jetbrains/java/decompiler/modules/decompiler/IfPatternMatchProcessor.java
Kroppeb: Yeah, I'm just not sure what to do with it

jaskarth: I think you can remove the `(Nothing is as permanent as a temporary solution :p)` and the `:/` at the end there, that'd make it okay with me

jaskarth (approved):
Looks good now!

joe455tech (approved):

jaskarth:
The latest commit fixes compilation issues within Minecraft, so I think it should be good to merge now.

PR #264, opened 2023-04-21
fix improper offset with generics

ix0rai:
this fix is needed for enigma, since I want to make Quiltflower the default decompiler and this was the last issue to track down first. I also considered moving line 196 out of its else block as a fix, but that seems like it would just be a slower way to accomplish the same thing from a quick look.

fixes https://github.com/QuiltMC/enigma/issues/47, which was causing some generics to explode and look like this:
image

all credit for tracking this down goes to my brilliant friend @orifu :)

### DIFF
--- a/src/org/jetbrains/java/decompiler/util/TextBuffer.java
+++ b/src/org/jetbrains/java/decompiler/util/TextBuffer.java
@@ -187,7 +187,9 @@ public class TextBuffer {
         if (gt.type == CodeConstants.TYPE_OBJECT) {
           addAllClassTokens(index + offset, name, gt.value);
         }
+
         addGenericTypeTokens(index + offset, name, gt);
+        offset += name.length();
       } else {
         if (t.type == CodeConstants.TYPE_OBJECT) {
           addTypeNameToken(t, index + offset);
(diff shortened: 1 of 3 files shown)


jaskarth:
Looks nice! Can you add a unit test for this?

ix0rai:
should I be worried about the currently failing tests?

sschr15:
No - QF's unit tests are cursed at best currently

ix0rai:
no idea if I did tests right, but I managed to make a test that fails without the change and succeeds with it

src/org/jetbrains/java/decompiler/util/TextBuffer.java
```diff
@@ -187,7 +187,9 @@ public TextBuffer addGenericTypeTokens(int index, String text, VarType type) {
         if (gt.type == CodeConstants.TYPE_OBJECT) {
           addAllClassTokens(index + offset, name, gt.value);
         }
+
         addGenericTypeTokens(index + offset, name, gt);
+        offset += name.length();
```
ByMartrixx, line 192: Huh, what an oversight (it was my fault)

ByMartrixx (approved):

OroArmor (approved):

sschr15 (approved):

jaskarth (approved):
LGTM!

PR #273, opened 2023-04-29
Add lazy library sources

Earthcomputer:
Add lazy library sources to avoid needing to list available classes in advance. This is the first step towards allowing the IntelliJ plugin to access libraries to improve decompilation output.

I tried my best to make this fit with the existing codebase, let me know if anything could be improved.


jaskarth:
High-level code review happened over Discord, but for posterity the plan is
to move the lazy tag of libraries away from the `Fernflower` and into the
context source directly.

jaskarth (approved):
Looks good, thanks so much for handling this!

PR #275, opened 2023-05-04
Kotlin property decompilation

sschr15:
This provides decompilation of most aspects of Kotlin properties, such as typing, flags, custom getters/setters, and delegates. It also provides more wrapper types for additional Kotlin sugaring.


plugins/kotlin/src/main/java/org/quiltmc/quiltflower/kotlin/KotlinWriter.java
```diff
@@ -125,6 +130,20 @@ private static boolean invokeProcessors(TextBuffer buffer, ClassNode node) {
...
+
+      List<KProperty> properties = new ArrayList<>();
+      Set<String> propertyFields = new HashSet<>();
+      Set<String> propertyAccessors = new HashSet<>();
+      boolean success = KProperty.parse(node, properties, propertyFields, propertyAccessors);
+      if (success) {
+        this.properties = properties;
+        this.propertyFields = propertyFields;
+        this.propertyAccessors = propertyAccessors;
+      } else {
```
jaskarth, line 142: I think this whole part with the lists and sets need to be refactored. Why not assign them all to final arraylists/hashsets, use the fields in the parse call, and if !success then clear the lists/sets? You would probably have to clear them beforehand too. Or, does this need shared state in the first place? Can this be refactored to use a param and a class?

plugins/kotlin/src/main/java/org/quiltmc/quiltflower/kotlin/KotlinWriter.java
```diff
@@ -1099,6 +1124,47 @@ else if (methodWrapper.varproc.getVarFinal(new VarVersionPair(index, 0)) == VarT
     return !hideMethod;
   }
 
+  public static boolean writeMethodBody(
+    ClassNode node,
+    MethodWrapper methodWrapper,
+    TextBuffer buffer,
+    int indent,
+    boolean hideIfInit
+  ) {
```
jaskarth, line 1133: Method def could be one line

plugins/kotlin/src/main/java/org/quiltmc/quiltflower/kotlin/struct/KProperty.java
```diff
@@ -0,0 +1,359 @@
...
+
+  @Nullable
+  public final StructField underlyingField;
+
+  @Nullable
+  public final Exprent initializer;
+
+  private final ClassesProcessor.ClassNode node;
+
+  public KProperty(
```
jaskarth, line 56: Constructors of this size should be private

plugins/kotlin/src/main/java/org/quiltmc/quiltflower/kotlin/struct/KProperty.java
```diff
@@ -0,0 +1,359 @@
...
+import org.quiltmc.quiltflower.kotlin.KotlinWriter;
+import org.quiltmc.quiltflower.kotlin.metadata.MetadataNameResolver;
+import org.quiltmc.quiltflower.kotlin.util.KTypes;
+import org.quiltmc.quiltflower.kotlin.util.ProtobufFlags;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+
+public class KProperty {
```
jaskarth, line 33: Would be better if it were final. Also, this class is rather complicated and so justifies the need for more code comments throughout.

plugins/kotlin/src/main/java/org/quiltmc/quiltflower/kotlin/struct/KProperty.java
```diff
@@ -0,0 +1,359 @@
...
+      ProtobufFlags.Property flags = new ProtobufFlags.Property(property.getFlags());
+
+      String name = nameResolver.resolve(property.getName());
+
+      String propDesc;
+      KType type;
+      if (property.hasReturnType() && property.getReturnType().hasClassName()) {
+        type = KType.from(property.getReturnType(), nameResolver);
+        propDesc = KTypes.getJavaSignature(type.kotlinType, property.getReturnType().getNullable());
+      } else {
```
jaskarth, line 262: No need for else statement, just assign variables to null ahead of time

plugins/kotlin/src/main/java/org/quiltmc/quiltflower/kotlin/struct/KType.java
```diff
@@ -0,0 +1,124 @@
+package org.quiltmc.quiltflower.kotlin.struct;
+
+import kotlin.reflect.jvm.internal.impl.metadata.ProtoBuf;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.java.decompiler.struct.gen.VarType;
+import org.jetbrains.java.decompiler.util.TextBuffer;
+import org.quiltmc.quiltflower.kotlin.metadata.MetadataNameResolver;
+import org.quiltmc.quiltflower.kotlin.util.KTypes;
+
+public class KType {
```
jaskarth, line 10: Also needs comments here. Also, is it possible to extend VarType? I would rather the type system be more uniform. See GenericType for more information.

jaskarth (changes requested):
Thanks for handling this! I have a couple of comments and requests, mostly about code style but some architecturally as well.

jaskarth (approved):
Looks good, thanks for doing this!

jaskarth:
Because of the impending merge of 1.9.x into 1.10 and the failing tests, I'll leave this for now and merge this once main has been merged into 1.10, and I'll fix the tests while doing so.

PR #289, opened 2023-07-16
add test for jad style naming param collision

MiniDigger:
the issue: if you have a param name that matches a generated jad style named local var, you get uncompilable code.

VF should take existing LVT into account when doing jad style naming

(idk how to best handle adding this test, the expected part is basically made up so that the test actually fails)


MiniDigger:
#291

PR #290, opened 2023-07-16
Support building on ARM64

zml2008:
(such as apple silicon)

The JVMs that are built for aarch64 are a bit more limited -- I've added the new gradle-recommended toolchain resolver which can find more JVMs, and introduced some logic to pick a 'next best' JVM for compiling test data when running on an ARM system. This allows a full build to run without issue hopefully!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,6 +18,7 @@ apply plugin: 'signing'
 
 allprojects {
   apply plugin: 'java'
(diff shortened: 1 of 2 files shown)


jaskarth (approved):
LGTM

jaskarth:
Thanks for this patch!

PR #291, opened 2023-07-17
Fix local variable jad name and type where the debug name conflicts

coehlrich:
Fixes cases in lambdas where the debug local variable name is used instead of the jad styled local variable name along with in certain cases the type of the local variable not being correct.


jaskarth (approved):
Ran it over the regression suite and got this result: https://gist.github.com/jaskarth/363ce347e7204eae63fc953035f8d6d4

Looks good to me.

PR #292, opened 2023-07-18
fix: consider method params when checking for clashing names

MiniDigger:
current test class:
```java
import java.util.Iterator;

public class TestJadLvtCollision implements Iterable<TestJadLvtCollision> {

    public void test(TestJadLvtCollision testjadlvtcollision) {
        TestJadLvtCollision testjadlvtcollision2 = new TestJadLvtCollision();
        System.out.println(testjadlvtcollision.toString() + testjadlvtcollision2.toString());
    }

    public void testLoop(TestJadLvtCollision testjadlvtcollision) {
        for (TestJadLvtCollision testjadlvtcollision2 : testjadlvtcollision) {
            System.out.println(testjadlvtcollision.toString() + testjadlvtcollision2.toString());
        }
    }

    @Override
    public Iterator<TestJadLvtCollision> iterator() {
        return null;
    }
}

```


MiniDigger:
found out that stuff still breaks in for-each loops, pushed a test, looking into it

jaskarth (approved):
There are some minor nitpicks, but not enough to block the PR. Thank you for this!

PR #295, opened 2023-07-19
Java version note on the ReadMe

drori200:
A small line change to add that to use VineFlower you need at least java 11


zml2008:
tbh this summary needs a more thorough review anyways, it hasn't been updated for the new Maven coordinates even.

jaskarth:
I concur- I think this would be best done in tandem with a broader readme refactor. Thanks for looking into this though, it is appreciated!

PR #303, opened 2023-08-20
fix: pull in processing decompile phase from Forgeflower threading impl

zml2008:
wip, needs to re-incorporate VF error handling to emit the appropriate info


zml2008:
(this will also need a backport to 1.9.3, but that shouldn't be too difficult after work is complete)

PR #306, opened 2023-09-05
Inline variables when doing pattern matching instanceof

coehlrich:
Inlines variables from before the if statement when doing pattern matching instanceof


jaskarth (approved):
Looks good, thank you!

PR #308, opened 2023-09-06
Add test for lambda param type loss

Juuxel:
VF currently fails to retain type information for the params here, leading to either
1. different behaviour since the lambda now accepts a wider range of inputs, silently eliminating ClassCastExceptions, or
2. code that doesn't compile since it tries to use the more specific parameter type, but fails.

The added test code has both cases.


jaskarth (approved):
Thanks for this test case!

PR #311, opened 2023-09-08
Fix final variable definition that is used in a for loop being moved outside the for loop

coehlrich:
Fixes a compile error in `RealmsSelectWorldTemplateScreen` where a variable is used in a lambda but the definition is being moved outside a for loop causing the variable to no longer be effectively final.


jaskarth (approved):
Patch looks good, but a unit test would be great to prevent regressions.

PR #313, opened 2023-09-09
Create some test cases for compile errors in decomp output

MiniDigger:


jaskarth (approved):
Thanks for these tests!

PR #318, opened 2023-09-10
Consider lambdas when searching where to place local class definitions (fixes #314)

### LINKED ISSUE #314: Local record is put at the end of the method, not the start
see TestLocalRecord#test5 (after #313 has been merged)

VF turns:
```java
  public CompletableFuture<Object> test5() {
    record R(Object a, Object b) {}
    return new CompletableFuture<>()
      .thenCombineAsync(null, R::new);
  }
```
into: 
```java
public CompletableFuture<Object> test5() {
      return new CompletableFuture().thenCombineAsync(null, (x$0, x$1) -> new R(x$0, x$1));// 35 36

      record R(Object a, Object b) {
      }

   }
```
which doesnt compile because R is referenced before its defined

MiniDigger:
I am not sure if this is good enough, but it fixes the test and doesn't cause a change in any other test, so it should be a good first step


jaskarth (approved):
Reasonable as an interim patch. Thank you!

PR #320, opened 2023-09-11
Make containsKey return true if the key is present and the value is null in NullableConcurrentHashMap

coehlrich:
`super.containsKey` calls `this.get(key)` and returns true if the value is not null which means if the key is in the map and the value is null then `containsKey` returns false.


jaskarth (approved):
Fantastic catch, thank you!

PR #321, opened 2023-09-11
Allow using generics with pattern matching instanceof

coehlrich:
Adds support for generics in the type part of pattern matching instanceof.


jaskarth (approved):
Simple fix, tested and looks good.

PR #322, opened 2023-09-12
Fix keeping generics that are unavailable in the current context

coehlrich:
Fixes some cases where generics were kept even though they aren't available.


jaskarth (approved):
Looks good, thank you!

PR #324, opened 2023-09-21
Add tool to extract CLI options

Ma5onic:
Added script to extract CLI options from the source code without having to compile the entire application


sschr15 (commented):
This is an interesting idea, but I'm not very sure if adding it to the VF repository is the best option. Also, the 1.10 versions have redone CLI arguments resulting in what would be a much cleaner extraction script with no need for a type table

zml2008:
I'm not particularly a fan of hand-parsing the source code -- it seems extremely brittle. See https://github.com/Vineflower/vineflower-site/pull/13 for another approach that I worked on a while ago, that's already fully integrated into our site.

Note that that too is not part of the main source tree -- I was originally thinking about putting the generator in the main repository, but most solutions are pretty specialized for a specific source format, and may want to update their output across a variety of VF versions, so I settled on keeping the generator as part of the site repository, where the output is actually used. I'd suggest similar for your script -- keep it with whatever is using the output.

PR #325, opened 2023-10-11
Fix potential exception by removing put outside computeIfAbsent

oicu0619:
this.classes is a ConcurrentHashMap. The originalCode try to put something into this.classes inside computeIfAbsent. computeIfAbsent will acquire lock of certain slot of the ConcurrentHashMap. If clazz.qualifiedName end up in the same slot, it will also try to accquire the same lock, which is already held by the same thread, will cause "java.lang.IllegalStateException: Recursive update" 
[demo.zip](https://github.com/Vineflower/vineflower/files/12865482/demo.zip)
This is a small jar file that will trigger the excetpion.


jaskarth (approved):
Thanks for this contribution, and sorry that it took so long for me to get back to you! The 1.10.0 dev branch has a fix for this, but this is a nice stop-gap fix. I will merge once I set up the dev branch for the 1.9.4 release.

jaskarth:
Hey! Sorry about not getting back sooner, but plans changed and 1.10.0 was released first, and 1.9.4 was skipped. Thanks for the contribution though! It is very much appreciated.

PR #327, opened 2023-11-10
Add machine-readable annotation to describe dynamic default values

shartte:
Add machine-readable annotation to describe dynamic default values such that DocGen can display a description of how that default value is computed instead of the default value used by the build agent building the documentation.

While at it: Fix description of nls Parameter (it didn't specify that it was an option and which setting caused LF vs CLRF).

I ran into this while trying to figure out what the default thread-count for Vineflower was (to try to find a minimum viable heap size). Turns out, it uses all CPUs by default, but the Website says `2`, since that happens to be the vCPU count of the machine building the docs :-)

p.s.: Next PR would be on the docs repo to make DocGen actually make use of the annotation.


jaskarth (approved):
This is a good change, thank you! I will merge this once I make a branch for a 1.9.4 release.

As a sidenote, the New Line Separator is [deprecated and obsoleted](https://github.com/Vineflower/vineflower/commit/30df62cc45affabd986f6ffa2e2340315266ad0a) in the 1.10.0 dev branch. It still makes sense to have it here, though.

jaskarth:
Sorry about the delay, and thank you for the contribution!

PR #330, opened 2023-11-16
Fix @Override annotations emitting twice

### LINKED ISSUE #329: Don't emit Override annotations if classfile already contains it
While in most cases Override annotations aren't found within the classfile due to it being retention level source, it is possible for classfile processing tools to emit them. In those cases, Vineflower's own Override annotation processing can mistakenly write it again, leading to two annotations being emitted, which is not legal. This can be worked around by setting the override-annotation option to false.

mudkipdev:
Closes #329.


jaskarth (approved):
This change looks good to me! Thank you for the contribution, and sorry about the delays!

PR #332, opened 2023-11-24
feature: comment lambdas with their corresponding synthetic method

### LINKED ISSUE #331: Add option to display names of lambda methods
This would be helpful in the case of doing bytecode editing, in knowing what methods lambda code is housed in.

tildejustin:
Resolves #331.
Personal concerns:
"Annotate" might be the wrong word to use for the option name
Maybe comment in front is not ideal
I don't know if / how to add tests for this :P

### DIFF
--- a/src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
+++ b/src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
@@ -387,6 +387,11 @@ public interface IFernflowerPreferences {
   @Type(Type.BOOLEAN)
   String REMOVE_IMPORTS = "remove-imports";
 
+  @Name("Annotate Lambdas")
+  @Description("Mark lambdas with their respective synthetic method")
+  @ShortName("ala")
+  String ANNOTATE_LAMBDAS = "annotate-lambdas";
+
   Map<String, Object> DEFAULTS = getDefaults();
 
   static Map<String, Object> getDefaults() {
@@ -458,6 +463,7 @@ public interface IFernflowerPreferences {
     defaults.put(FORCE_JSR_INLINE, "0");
     defaults.put(DUMP_TEXT_TOKENS, "0");
     defaults.put(REMOVE_IMPORTS, "0");
+    defaults.put(ANNOTATE_LAMBDAS, "0");
 
     return Collections.unmodifiableMap(defaults);
   }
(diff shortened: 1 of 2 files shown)


tildejustin:
Should the anonymous class naming be under a different option?

SpaceWalkerRS:
if you're adding support for anonymous classes, then consider support for local classes as well

jaskarth:
I think the anonymous class naming can go under the same option. There is significant overlap between the two and I can't really think of cases where you might want one but not the other.

jaskarth (approved):
Looks good! Thanks for the contribution!

src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
```diff
@@ -387,6 +387,11 @@ public interface IFernflowerPreferences {
   @Type(Type.BOOLEAN)
   String REMOVE_IMPORTS = "remove-imports";
 
+  @Name("Mark Corresponding Sythetics")
```
Juuxel, line 390: Typo: "sythetics"

test/org/jetbrains/java/decompiler/SingleClassesTest.java
```diff
@@ -141,6 +141,14 @@ public String getMethodDoc(StructClass structClass, StructMethod structMethod) {
       IFernflowerPreferences.FORCE_JSR_INLINE, "1",
       IFernflowerPreferences.PREFERRED_LINE_LENGTH, "120"
     );
+    registerSet("Sythetics Marking", this::registerSytheticsMarking,
```
Juuxel, line 144: Typo: "sythetics" (in the string)

test/org/jetbrains/java/decompiler/SingleClassesTest.java
```diff
@@ -831,4 +839,10 @@ private void registerTextTokens() {
     register(JAVA_16, "TestTextTokens2");
     register(JAVA_8, "TestTextTokens3");
   }
+
+  private void registerSytheticsMarking() {
```
Juuxel, line 843: Typo: "sythetics"

src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
tildejustin: aah sorry, I caught myself making that typo a couple times but it wasn't enough apparently

PR #337, opened 2023-12-26
add the ability to display the method name for lambda's in a comment

wagyourtail:
Probably useful for making things like writing mixins targeting lambdas slightly easie


wagyourtail:
oh, looks like someone's basically already done this for .10

PR #339, opened 2024-01-21
Fix a multitude of crashes and exceptions during Kotlin decompilation

sschr15:
I [have run](https://gist.github.com/sschr15/e64a894a0b22b4b74c4e31eb7a9a8630) a diff against the decompilation of Kotlin's standard library on version 1.9.20 comparing these fixes against the current 1.10 commit


jaskarth (approved):
This looks reasonable to me. Thanks!

PR #341, opened 2024-01-26
Don't write explicit casts where redundant

IotaBread:


jaskarth (approved):
Looks good! Thanks for the PR :)

PR #346, opened 2024-02-14
fix a thing in the readme i just noticed

thisisverycool:


zlataovce:
VF is a continuation of Quiltflower, what's written there is correct

jaskarth:
This change is invalid.

PR #348, opened 2024-03-14
Add option to exclude classes from decompilation

mvisat:
Sometimes it's unnecessary to decompile certain classes, for example libraries such as Apache Commons. Skipping the decompilation of these classes can speed up the overall decompilation time. Currently there is `--only` option, but it doesn't cover cases where we don't know which classes in the JAR to decompile.

This PR adds `--exclude` or `-E` option to the decompiler. For convenience, it supports excluding multiple classes without multiple `--exclude` or `-E` invocations (can be separated by commas).

Examples:
```sh
# using --exclude to exclude org.apache.*
java -jar vineflower.jar --exclude=org/apache input.jar out_dir

# using -E to exclude multiple classes
java -jar vineflower.jar -E=org/apache,org/bouncycastle,com/google/android input.jar out_dir
```


Geolykt:
I believe the consensus is to move away from the "short" options, so I am a bit torn about them being introduced for a new option.

Additionally, I'd prefer if the option is added to <https://github.com/mvisat/vineflower/blob/49d8fd22f33df753aa2277d68adb548e0168151b/src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java>. At this point in time this is not done and as such I myself would be considering the option to not exist unless I were to know otherwise (that is all options that are outside the IFernflowerPreferences are undocumented to me).

mvisat:
Thanks for the feedback :) I've updated the code to use `IFernflowerPreferences`. The option is now (tentatively) `--excluded-classes` or `-exc`.

jaskarth:
Thanks a lot for this! I think the implementation is good, but my one request is to make the exclusion regex based rather than prefix based. It would help a lot for debugging.

jaskarth:
Hey! I'm really sorry, but while doing branch cleanup I accidentally deleted the 1.10 branch before changing the base to 1.11. If you are able to re-create the pull request targeting 1.11, that would be greatly appreciated. Sorry for the trouble!

PR #355, opened 2024-04-05
Fix switches on enums with a switch map along with null cases for java 21

coehlrich:
For enums in other classes a switch map is used and for switches with a null case pattern matching is used (which has also been released).


jaskarth (approved):
Looks good, thanks!

PR #358, opened 2024-04-06
Fix switch expressions not generating

coehlrich:
Fixes switch expressions not being converted from a switch statement when a case contains an if statement due to implicit edges created by the if statement along with switch statements that have a case that throws an exception instead of returning a value.

### DIFF
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/TryWithResourcesProcessor.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/TryWithResourcesProcessor.java
@@ -7,8 +7,11 @@ import org.jetbrains.java.decompiler.modules.decompiler.stats.*;
 import org.jetbrains.java.decompiler.struct.gen.VarType;
 
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


src/org/jetbrains/java/decompiler/modules/decompiler/TryWithResourcesProcessor.java
```diff
@@ -348,14 +351,21 @@ public static void findEdgesLeaving(Statement curr, Statement check, Set<StatEdg
...
 
   public static void findEdgesLeaving(Statement curr, Statement check, Set<StatEdge> edges, boolean allowExit) {
+    findEdgesLeaving(curr, check, edges, allowExit, new HashMap<>());
+  }
+
+  private static void findEdgesLeaving(Statement curr, Statement check, Set<StatEdge> edges, boolean allowExit, Map<Statement, Set<Statement>> found) {
     for (StatEdge edge : curr.getAllSuccessorEdges()) {
       if (!check.containsStatement(edge.getDestination()) && (allowExit || !(edge.getDestination() instanceof DummyExitStatement))) {
-        edges.add(edge);
+        if (edge.explicit || found.entrySet().stream().allMatch(e -> !e.getKey().containsStatement(curr) || !e.getValue().contains(edge.getDestination()))) {
```
jaskarth, line 360: If you could add a comment here as to what the check is doing, that'd be great :)

jaskarth (approved):
Looks good, thanks!

PR #361, opened 2024-04-07
Fix variable namer parent not being set for local classes

coehlrich:
Local classes are only added to the method body after parent contexts are added.


jaskarth (approved):
Looks good now! Thanks!

PR #369, opened 2024-04-12
Kotlin default arguments

sschr15:


plugins/kotlin/src/main/java/org/vineflower/kotlin/struct/DefaultArgsMap.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  public static DefaultArgsMap from(MethodWrapper defaults, MethodWrapper calling, KParameter[] params) {
+    if (defaults == null) {
+      return new DefaultArgsMap(Map.of());
+    }
+
+    Map<KParameter, Exprent> map = new HashMap<>();
+
+    DirectGraph graph = defaults.getOrBuildGraph();
```
jaskarth, line 31: This piece of code is significantly nontrivial. I think you should add some comments explaining what's going on here.

plugins/kotlin/src/main/java/org/vineflower/kotlin/struct/DefaultArgsMap.java
```diff
@@ -0,0 +1,101 @@
...
+    for (DirectNode node : graph.nodes) {
+      Statement statement = node.statement;
+      if (statement instanceof IfStatement ifStatement) {
+        Exprent condition = ifStatement.getHeadexprent().getCondition();
+        if (!(condition instanceof FunctionExprent function)) continue;
```
jaskarth, line 36: Avoid single line ifs, use ifs with braces instead.

jaskarth (changes requested):

jaskarth (approved):
Looks nice!

PR #370, opened 2024-04-14
Add option to exclude classes from decompilation with regex

mvisat:
Sometimes it's unnecessary to decompile certain classes, for example libraries such as Apache Commons. Skipping the decompilation of these classes can speed up the overall decompilation time. Currently there is `--only` option, but it doesn't cover cases where we don't know which classes in the JAR to decompile.

This PR adds that option to the decompiler (currently it's `--excluded-class-regex` or `-exc` for shorthand). It will skip the decompilation if the specified regex matches the fully qualified name of the class.

Examples:
```sh
# exclude all classes in org.apache package
java -jar vineflower.jar '-exc=org/apache/.*' input.jar out_dir

# exclude classes from multiple packages
java -jar vineflower.jar '-exc=org/(apache|bouncycastle)/.*|com/google/android/.*' input.jar out_dir
```

### DIFF
--- a/src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
+++ b/src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
@@ -385,6 +385,7 @@ public interface IFernflowerPreferences {
   @Name("Excluded Class Regex")
   @Description("Exclude classes from decompilation if their fully qualified names match the specified regular expression.")
   @ShortName("exc")
+  @Type(Type.STRING)
   String EXCLUDED_CLASS_REGEX = "excluded-class-regex";
 
   Map<String, Object> DEFAULTS = getDefaults();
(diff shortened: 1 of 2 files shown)


src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
```diff
@@ -382,6 +382,12 @@ public interface IFernflowerPreferences {
   @ShortName("mcs")
   String MARK_CORRESPONDING_SYNTHETICS = "mark-corresponding-synthetics";
 
+  @Name("Excluded Class Regex")
+  @Description("Exclude classes from decompilation if their fully qualified names match the specified regular expression.")
+  @ShortName("exc")
+  @Type(Type.STRING)
+  String EXCLUDED_CLASS_REGEX = "excluded-class-regex";
```
jaskarth, line 389: Thank you for making it a regex! I just have two small things before merging. For one, I think instead of `excluded-class-regex` the option should be named `exclude-classes`, since the regex is already specified in the description. `@ShortName("exc")` should also be removed since short names are legacy and no longer being added for new options. Thanks a lot!

mvisat: no problem, updated! :)

jaskarth (approved):
Looks good, thank you!

PR #375, opened 2024-04-17
Fix 2 constructors of a non-static inner class with 1 calling another

coehlrich:
Adds an `assigned` field to `VarFieldPair` so that when checking the fields they are only compared if they are known to be correct.


jaskarth (approved):
Looks good, thanks!

PR #376, opened 2024-04-17
Update Kotlin plugin to use modern Java features

sschr15:
Also fixes a couple issues involving constructor parameter defaults, including a full decompilation error.


jaskarth (approved):
Looks fine to me. Thanks!

PR #378, opened 2024-04-19
Fixes for clashing variable renaming

jpenilla:
This seems to have gotten lost in 6362c9408e5020814a35d4d669e0e4994680476e


jaskarth (approved):
Sorry about the very delayed update! This patch looks good to me, will merge. Thank you for fixing this!

PR #381, opened 2024-04-19
Simplify empty anonymous class bodies to `{}`

jpenilla:
This is typically how people actually write code like this (as can be seen in the existing tests).

The improvement is extra noticeable with things like type tokens.

I can add an option for this if wanted. It would also not be hard to expand this to other types of classes, but that seemed less suitable, at least as a default behavior.


jaskarth (changes requested):

jaskarth (approved):
Looks good now, thanks for the update :)

PR #384, opened 2024-04-26
Fix breaking method decompilation on switch-on-enum when enum does not exist in classpath

EpicPlayerA10:
Fixes breaking method decompilation on switch-on-enum when the enum does not exist in classpath. This usually happens when you don't add external jar to classpath.

Previously it wouldn't decompile any code in method, but now it will only notify that simplifying switch-on-enum failed.

Also added tests to make sure that it will decompile correctly when enum does not exist in classpath. The test must be in compiled `.class` because we must remove the enum from compiled class for testing purpose.


src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
```diff
@@ -313,14 +317,18 @@ private static boolean simplifyNewEnumSwitch(SwitchStatement switchSt) {
     if (inner instanceof InvocationExprent && ((InvocationExprent) inner).getName().equals("ordinal")) {
       InvocationExprent invInner = (InvocationExprent) inner;
 
-      ClassesProcessor.ClassNode classNode = DecompilerContext.getClassProcessor().getMapRootClasses().get(invInner.getClassname());
+      StructClass classStruct = DecompilerContext.getStructContext().getClass(invInner.getClassname());
+      if (classStruct == null) {
+        root.addComment("$VF: Unable to simplify switch-on-enum. You are probably missing external library", true);
```
Geolykt, line 322: nitpick: Probably should be "... missing **an** external library"

EpicPlayerA10: Fixed

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
```diff
@@ -313,14 +316,18 @@ private static boolean simplifyNewEnumSwitch(SwitchStatement switchSt) {
     if (inner instanceof InvocationExprent && ((InvocationExprent) inner).getName().equals("ordinal")) {
       InvocationExprent invInner = (InvocationExprent) inner;
 
-      ClassesProcessor.ClassNode classNode = DecompilerContext.getClassProcessor().getMapRootClasses().get(invInner.getClassname());
+      StructClass classStruct = DecompilerContext.getStructContext().getClass(invInner.getClassname());
+      if (classStruct == null) {
+        root.addComment("$VF: Unable to simplify switch-on-enum. You are probably missing an external library", true);
```
jaskarth, line 321: I think it'd be best to leave the `root.addComment` line out of the PR, since it will only be written for Java 21+ class files. In addition, I think the phrasing should be different as instead of mentioning that an external library might be missing (which could be incorrect!) we should say something along the lines of: `$VF: Unable to simplify switch-on-enum, as the enum class was not able to be found.` This would be better done in a different PR though, keeping this one purely a bugfix.

jaskarth (changes requested):
Hi, thanks a lot for this PR! I've included two lines of feedback which I think should be addressed, but otherwise this fix looks good to me.

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
```diff
@@ -285,7 +285,10 @@ private static boolean simplify(SwitchStatement switchStatement, StructMethod mt
     return false;
   }
 
-  private static boolean simplifyNewEnumSwitch(SwitchStatement switchSt) {
+  /**
+   * Simplifies the switch-on-enum statement. This is new feature in Java 21
```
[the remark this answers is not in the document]
EpicPlayerA10: Fixed

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
EpicPlayerA10: Why in different PR? This comment is relevant to this bugfix. We need to notify the user that vineflower detected this switch-on-enum form and something failed. Anyway, changed the message to your proposal.

jaskarth: My point is that it should warn for the other forms too, not just the J21+ one. I suppose it's fine though; I can take a look at that one at a later time.

EpicPlayerA10:
Rebased onto `develop/1.11.0`

jaskarth (approved):
Thanks for the rebase, looks good now!

PR #386, opened 2024-04-30
Map all bytecode in range

jaskarth:
Fixes many bytecode mapping bugs. PR since >300 tests are updated.


jaskarth:
Thanks for the review! Will merge now.

PR #389, opened 2024-05-04
Refactor the way bytecode offsets are tracked.

Kroppeb:
* Instructions now track their own bytecode offsets.
* Split `SimpleInstructionSequence` and `FullInstructionSequence`. They aren't used in the same places and very little logic was actually shared.
* Renamed `SimpleInstructionSequence` to `InstructionSequence`
* Remove usages of `VBStyleCollection`
* Simplified and converted `ExceptionHandler` into a record


src/org/jetbrains/java/decompiler/struct/StructMethod.java
```diff
@@ -319,41 +320,40 @@ else if (opcode >= opc_invokevirtual && opcode <= opc_invokestatic) {
...
+      } else {
+        exceptionClass = pool.getPrimitiveConstant(excClass).getString();
       }
 
-      lstHandlers.add(handler);
+      lstHandlers.add(new ExceptionHandler(from, to, handler, exceptionClass));
     }
 
-    InstructionSequence seq = new FullInstructionSequence(instructions, new ExceptionTable(lstHandlers));
+    FullInstructionSequence seq = new FullInstructionSequence(
```
jaskarth, line 350: Move back to one line

jaskarth (changes requested):
I think this generally looks good but since this code is being significantly rewritten `FullInstructionSequence` and `InstructionSequence` need top level comments explaining what they are and how they're used.

Kroppeb:
So many conflicts :panic:

jaskarth (approved):
Looks good, I merged it manually due to the commit mess that I caused earlier.

PR #390, opened 2024-05-05
Fix for `null, default` switch case not being last

jpenilla:
In some cases after different processing stages, the default case is no longer last. This reorders it to last if necessary after all other switch processing.

A better solution may be to use the last instead of the first index when merging edges, this would likely improve the accuracy of TestAssignmentSwitchExpression3 output (as in case order matching input source).


jaskarth:
Closing this one since I merged the other one instead. But thanks a lot for looking into this issue!

PR #391, opened 2024-05-05
Fix default case not always being last in switch statements

coehlrich:
reorders the cases so that the default case is always last because if a case pattern or null literal is used then the default case (if present) must be last to successfully compile.


jaskarth (approved):
The fix looks good, thanks a lot!

PR #392, opened 2024-05-06
Add option to disable inner classes names validation

EpicPlayerA10:
Added option to disable inner classes names validation. Useful when the jar that you are decompiling does not follow inner classes naming convention (i.e. does not have `$` in inner class names).


jaskarth (approved):
Looks good. Thanks for the PR, and sorry for the very delayed review!

Issue #388, opened 2024-05-03
Some of the removed redundant casts aren't so redundant after all


Machine-Maker:
## Vineflower version
1.11.0-20240501.032551-24

## Describe the bug
Previous versions of VF had casts to long in places that are required for Minecraft source to run. It compiles, but errors on startup.

## Additional information
```
Index: net/minecraft/world/level/levelgen/BitRandomSource.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/net/minecraft/world/level/levelgen/BitRandomSource.java b/net/minecraft/world/level/levelgen/BitRandomSource.java
--- a/net/minecraft/world/level/levelgen/BitRandomSource.java    (revision 945cfbb91d450971c7076db24b5a76a4ea97cd0c)
+++ b/net/minecraft/world/level/levelgen/BitRandomSource.java    (date 1714775725817)
@@ -18,7 +18,7 @@
         if (bound <= 0) {
             throw new IllegalArgumentException("Bound must be positive");
         } else if ((bound & bound - 1) == 0) {
-            return (int)(bound * this.next(31) >> 31);
+            return (int)((long)bound * (long)this.next(31) >> 31);
         } else {
             int i;
             int i1;
```
this is a fix required with the above version of VF in order to make the game run. (re-adding the (long) casts).

Machine-Maker:
I just added a small thing to the tests that now fail which show the issue more directly. https://github.com/Machine-Maker/vineflower/tree/demo/cast-issue

jaskarth:
Fixed by https://github.com/Vineflower/vineflower/pull/395, thanks for the report and fix!

PR #395, opened 2024-05-16
Add back casts in numeric operations where it increases the size of the number

Machine-Maker:
Fixes https://github.com/Vineflower/vineflower/issues/388

Also lets decompiled minecraft actually start again.


jaskarth (approved):
Looks good, thanks a lot for the fix!

PR #399, opened 2024-05-20
Fix variable used in switch head not being inlined

coehlrich:
Fixes cases where vineflower creates an unitialized variable and checks if it's non null then casting that unitialized variable then putting the exprent that should be in the switch head above the switch statement with the variable in the switch head being the casted null variable.

### DIFF
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/SwitchPatternMatchProcessor.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/SwitchPatternMatchProcessor.java
@@ -5,6 +5,7 @@ import org.jetbrains.java.decompiler.main.DecompilerContext;
 import org.jetbrains.java.decompiler.main.collectors.CounterContainer;
 import org.jetbrains.java.decompiler.main.extern.IFernflowerPreferences;
 import org.jetbrains.java.decompiler.modules.decompiler.exps.*;
+import org.jetbrains.java.decompiler.modules.decompiler.exps.FunctionExprent.FunctionType;
 import org.jetbrains.java.decompiler.modules.decompiler.stats.*;
 import org.jetbrains.java.decompiler.struct.consts.PooledConstant;
 import org.jetbrains.java.decompiler.struct.consts.PrimitiveConstant;
@@ -15,6 +16,7 @@ import org.jetbrains.java.decompiler.struct.gen.VarType;
 import org.jetbrains.java.decompiler.util.Pair;
 
 import java.util.*;
+import java.util.stream.Stream;
 
 public final class SwitchPatternMatchProcessor {
   public static boolean processPatternMatching(Statement root) {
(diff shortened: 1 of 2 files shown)


src/org/jetbrains/java/decompiler/modules/decompiler/SwitchPatternMatchProcessor.java
```diff
@@ -262,6 +266,41 @@ private static boolean processStatement(SwitchStatement stat, Statement root) {
...
+              basicHead.remove(basicHead.size() - 1);
+            }
+          }
+        }
+      }
+    }
+
+    if (oldSelector != realSelector) {
+      Exprent finalSelector = realSelector;
+      stat.getCaseValues().stream().flatMap(List::stream).filter(Objects::nonNull).filter(exp -> exp instanceof FunctionExprent func && func.getFuncType() == FunctionType.INSTANCEOF && func.getLstOperands().get(0).equals(oldSelector)).forEach(exp -> ((FunctionExprent) exp).getLstOperands().set(0, finalSelector));
```
jaskarth, line 301: This line needs to be broken up into multiple lines, and it needs a comment explaining what it's trying to do.

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchPatternMatchProcessor.java
```diff
@@ -262,6 +266,41 @@ private static boolean processStatement(SwitchStatement stat, Statement root) {
       }
     }
 
+    Exprent oldSelector = realSelector;
+    // inline head
+    List<Exprent> basicHead = stat.getBasichead().getExprents();
+    if (realSelector instanceof VarExprent var && basicHead != null && basicHead.size() >= 1) {
+      if (basicHead.get(basicHead.size() - 1) instanceof AssignmentExprent assignment && assignment.getLeft() instanceof VarExprent assigned) {
+        if (var.equals(assigned) && !var.isVarReferenced(root, Stream.concat(Stream.of(assigned), stat.getCaseValues().stream().flatMap(List::stream).filter(exp -> exp instanceof FunctionExprent func && func.getFuncType() == FunctionType.INSTANCEOF && func.getLstOperands().get(0) instanceof VarExprent checked && checked.equals(var)).map(exp -> (VarExprent) ((FunctionExprent) exp).getLstOperands().get(0))).toArray(VarExprent[]::new))) {
```
jaskarth, line 274: Stream should be broken up into multiple lines

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchPatternMatchProcessor.java
```diff
@@ -262,6 +266,41 @@ private static boolean processStatement(SwitchStatement stat, Statement root) {
       }
     }
 
+    Exprent oldSelector = realSelector;
```
jaskarth, line 269: Since this is a large amount of code, it should have a heading describing what it's trying to do. Such as:

```suggestion
    // Try to inline:
    // Objects.requireNonNull(var);
    // switch (var) {
    // 
    // to
    // switch (var) {
    
    Exprent oldSelector = realSelector;
```

jaskarth (changes requested):
Thanks a lot for this change! The fix itself looks fine to me, I have mostly stylistic changes here.

jaskarth (approved):
Looks good, thanks!

PR #401, opened 2024-05-23
Fix stack size of first non synthetic parameter being used for synthetic parameters

coehlrich:
Fixes a case where it would always use the stack size of the first real parameter for synthetic parameters if the method has a signature


jaskarth (approved):
Looks good, thanks!

PR #405, opened 2024-06-22
Add support for intersection types in casts

coehlrich:
Adds support for intersection types in casts when used for a method or variable.


jaskarth (approved):
This looks good to me. Thanks a lot for the work!

PR #414, opened 2024-08-17
Fix local record classes that have the parent class as the first parameter

### LINKED ISSUE #413: Local record constructors are incorrect
## Vineflower version
1.10.1

Given the source code:
```java
public class TestLocalRecord {
        public Object test(List<Integer> list) {
                record Rec(TestLocalRecord a, List<Integer> b) {};

                return new Rec(this, list);
        }
}
```

The output is:
```java
public class TestLocalRecord {
   public Object test(List<Integer> var1) {
      record Rec(TestLocalRecord a, List<Integer> b) {
         Rec(TestLocalRecord b) {
            this.b = b;
         }
      }

      return new Rec(var1);
   }
}
```
The extra constructor is incorrect.

coehlrich:
Fixes #413 

https://docs.oracle.com/javase/specs/jls/se16/html/jls-8.html#jls-8.10
> A nested record class is implicitly static.


jaskarth (approved):
Looks good, thanks for the fix!

PR #415, opened 2024-08-18
Remove final from `Statement` subtypes

sschr15:
To better work with plugins, removing the final modifier from `Statement` subclasses would improve the experience ([feature/kotlin/loop-changes](https://github.com/Vineflower/vineflower/tree/feature/kotlin/loop-changes) would subclass statements if available)


jaskarth (commented):
I think mostly this is fine, but I would prefer `BasicBlockStatement`, `RootStatement`, `DummyExitStatement`, and `GeneralStatement` to remain final because these are hardcoded in many places with explicitly defined meanings.

jaskarth (approved):
Thanks!

PR #417, opened 2024-08-18
Adds support for nested record patterns and record patterns in switches

coehlrich:


src/org/jetbrains/java/decompiler/modules/decompiler/SwitchPatternMatchProcessor.java
```diff
@@ -345,6 +365,23 @@ private static boolean eliminateGuardRef(SwitchStatement stat, Map<List<Exprent>
     // alternatively, it can be inverted as `if (guardCond) { /* regular case code... */ break; } idx = __thisIdx + 1;`
     if (reference.b instanceof AssignmentExprent) {
       Statement assignStat = reference.a;
+      int statementIndex = -1;
```
jaskarth, line 368: Needs whitespace after `}` for the for loop and while loop, and a comment explaining what this block of code is doing

jaskarth (approved):
Looks good otherwise, just one review note from me.

PR #418, opened 2024-08-18
Decompiler option access API

sschr15:
This also moves `ConsoleHelp` to use the API class instead of re-parsing again.


jaskarth (changes requested):
I think `getAll` and `getAllByPlugin` need javadoc, but other than that this looks good!

jaskarth (approved):
Looks good, thanks!

PR #420, opened 2024-08-21
Allow plugins to change decompiled file extensions

sschr15:


src/org/jetbrains/java/decompiler/api/plugin/LanguageSpec.java
```diff
@@ -8,12 +8,18 @@ public final class LanguageSpec {
   public final GraphParser graphParser;
   public final StatementWriter writer;
   public final Pass pass;
-
+  public final String extension;
+  
   public LanguageSpec(String name, LanguageChooser chooser, GraphParser graphParser, StatementWriter writer, Pass pass) {
+    this(name, chooser, graphParser, writer, pass, "java");
```
jaskarth, line 14: I think this should be removed. It's technically a breaking change but I don't think anyone else has written language plugins yet and defaulting to java is not ideal.

testFixtures/org/jetbrains/java/decompiler/SingleClassesTestBase.java
```diff
@@ -247,14 +268,14 @@ public void run(Object[] options, SingleClassesTestBase base) throws IOException
           throw new UncheckedIOException(e);
         }
         //noinspection ConstantConditions
-        assumeTrue(false, referenceFile.getFileName() + " was not present yet");
+        abort(referenceFile.getFileName() + " was not present yet");
```
jaskarth, line 271: What's the reasoning behind this change?

testFixtures/org/jetbrains/java/decompiler/SingleClassesTestBase.java
```diff
@@ -220,16 +228,29 @@ public void run(Object[] options, SingleClassesTestBase base) throws IOException
 
       String testFileName = classFile.getFileName().toString();
       String testName = testFileName.substring(0, testFileName.length() - 6);
-      Path decompiledFile = fixture.getTargetDir().resolve(testName + ".java");
-      assertTrue(Files.isRegularFile(decompiledFile));
+
+      Path decompiledFile;
+      if (expectedFileName == null) {
```
jaskarth, line 233: I think this logic is fairly complex now, it needs some comments.

jaskarth (changes requested):
Thanks for doing this! I've added some comments here.

testFixtures/org/jetbrains/java/decompiler/SingleClassesTestBase.java
sschr15: The reasoning was to follow the `fail` code I had above, but I've reverted that with my new commit

jaskarth (approved):
Looks good now, thanks!

PR #423, opened 2024-08-24
Fix lambdas in anonymous classes referencing parameters from the root method

coehlrich:
Fixes cases where a lambda in an anonymous class that is in a method references a parameter from the root method.


Machine-Maker (approved):

jaskarth (approved):
This is fantastic, thank you!

PR #430, opened 2024-09-14
Add checks for generic types and discard any unexpected types

### LINKED ISSUE #409: Corrupt generic signatures lead to incorrect class super types/interfaces, field types, and method parameter/variable types
## Vineflower version
1.10.1

## Describe the bug
If generic signatures are present, Vineflower replaces the corresponding types with those in the signatures, without verifying that they are the same.

## Additional information
For example, take the following class

```java
import java.util.ArrayList;
import java.util.List;

abstract class Signatures implements List<String> {
	List<String> field = new ArrayList<>();

	void method(List<String> o) {
	}
}
```

The class `Signatures` has interface `java/util/List` and signature `Ljava/lang/Object;Ljava/util/List<Ljava/lang/String;>;`.
The field `field` has descriptor `Ljava/util/List;` and signature `Ljava/util/List<Ljava/lang/String;>;`.
The method `method` has descriptor `(Ljava/util/List;)V` and signature `Ljava/util/List<Ljava/lang/String;>;`.

Now if we corrupt the class file by changing only the signatures, as follows:
Change the class signature to `Ljava/lang/Object;Ljava/util/Map<Ljava/lang/String;Ljava/lang/String;>;`.
Change the field signature to `Ljava/util/Map<Ljava/lang/String;Ljava/lang/String;>;`.
Change the method signature to `(Ljava/util/Map<Ljava/lang/String;Ljava/lang/String;>;)V`.

This corrupted class file will decompile to

```java
import java.util.ArrayList;
import java.util.List;
import java.util.Map;

abstract class Signatures implements Map<String, String> {
	Map<String, String> field = new ArrayList();

	void method(Map<String, String> o) {
	}
}
```

Notice that the class interface, field descriptor, and method descriptor all changed, while those were not modified in the class file at all.

In situations like this Vineflower ought to ignore corrupted signatures rather than corrupting other code with it.

Here's the class files I tested with:
[uncorrupted class file](https://github.com/user-attachments/files/16367608/signatures.zip)
[corrupted class file](https://github.com/user-attachments/files/16367609/signatures-corrupted.zip)

sschr15:
Fixes #409 by adding checks and discarding any clearly bogus generics


SpaceWalkerRS:
How do these checks work when signatures depend on each other? For example, a field that initially has a valid signature, but its type class's signature does not pass the check and is modified/removed, which in turn then invalidates that field's signature.

jaskarth (approved):
Looks fine, thanks!

PR #436, opened 2024-11-02
Fix requireNonNull call with reusable selectors and fix default being removed from exhaustive switch statements

coehlrich:
fixes `Objects.requireNonNull` not being removed when the selector for the switch statement is reusable and fixes the default case value being removed from switches that are required to be exhaustive.


jaskarth (approved):
Thanks for the work, and sorry for the delay! This looks good to me.

PR #437, opened 2024-11-02
Add test for labeled blocks in constructor

Earthcomputer:
This incorrectly decompiles to a `super()` call inside of the labelled block, add a test so it can be fixed later


jaskarth (approved):
Thanks for the unit test! This looks good to me.

PR #443, opened 2024-12-07
Mark config item `mcs` as a bool

Ecdcaeb:
Mark config item `mcs` as a bool
Currently it lacks a type annotation and is ignored by `writeOptions`.


Ecdcaeb:
X.X

zml2008:
hm?

Ecdcaeb:
I got the wrong base. Is this mergable?

zml2008:
No reason it shouldn't be with a bit of fiddling

PR #446, opened 2024-12-16
emit right variable name for catch variables, fixes #445

### LINKED ISSUE #445: lvt names for the exceptions in catch clauses aren't emitted
## Vineflower version
1.11 @ ef138de8152ddf0f3f1b8cb7505d5ca35737af61

## Describe the bug
right now var names are just auto incremented here https://github.com/Vineflower/vineflower/blob/master/src/org/jetbrains/java/decompiler/modules/decompiler/stats/CatchStatement.java#L50, I expect to see the name listed in the lvt.

## Additional information
https://github.com/Vineflower/vineflower/blob/master/testData/src/java8/pkg/TestTryCatchFinally.java#L28 is emitted as var8, not e2
https://github.com/Vineflower/vineflower/blob/master/testData/results/pkg/TestTryCatchFinally.dec#L10

from discord:
> i believe it's because the variables the catch statement adds are synthetic
> basically to fix it i think the catch variables need to be associated with lvts
~ Jasmine

MiniDigger:
draft since there are a whole bunch of failing tests, I only updated expect where I think the output is right


jaskarth:
Closing since a different implementation was developed. Thanks for the work, though!

PR #449, opened 2024-12-22
Add test for Expression Optimization Issue

Owen1212055:
This includes a test for a strange issue that causes LVT reassignments to be lost. 

## Source:
```java
double one = 1;
double shouldBeOne = one;
one = 0;
if (one > 1) {
}

System.out.println(shouldBeOne); // -> 1.0
```

## Decompiles to:
```java
double one = 1.0;
one = 0.0;
if (one > 1.0) {
}

System.out.println(one); // -> 0.0
```

### Odd Case
It should be noted that without the if statement, it appears to work fine.
```
// Decompiles to without if statement
double one = 1.0;
double var5 = 0.0; // <- although interestingly, this LVT name is lost
System.out.println(one);
```


jaskarth (approved):
Thanks for the unit test! It looks good to me.

PR #450, opened 2025-01-08
Fix pattern matching instanceof not being detected inside loops

coehlrich:
Fixes pattern matching instanceof not always being detected inside loops due to the predecessors check ending up inside the statement


jaskarth (approved):
Looks good, thanks for the fix!

PR #451, opened 2025-01-10
Fix incorrect ppmm inlining

Kroppeb:
Fixes that `foo(++a, a)` gets decompiled to `foo(a, ++a)` and all other related issues.


jaskarth (approved):
Looks good, thank you!

PR #461, opened 2025-02-26
Prevent NPE on #getRecordComponents for Object

lynxplay:
Requesting the record components from a struct class has a fallback to
an empty list over null for objects that extend j.l.Record but do not
have a record attribute.

This fallback however assumes that the type has a superclass, which is
not the case for j.l.Object, causing a NPE.


Earthcomputer (approved):

lynxplay:
Fixed commit type :tm:

jaskarth (approved):
Looks good, thanks!

PR #463, opened 2025-02-26
Stricter type pattern matching rules

lynxplay:
Vineflower's previous implementation of the type casting pattern
matching in switch statements was not strict enough, considering any
form of cast in the first expression of the case block a result of type
pattern matching.

This would incorrectly attempt to emit such type pattern matching source
code for switches that simply yielded different types.

This commit tightens the logic to consider a switch case a type pattern
matching one if the first expression is a simple cast of the switch
head's variable.


jaskarth (approved):
Looks good, thanks a lot!

PR #464, opened 2025-02-27
Fix ambiguous method calls with varargs that are numbers

coehlrich:


jaskarth (approved):
Looks good, thanks for fixing!

PR #471, opened 2025-03-22
Delegate semicolons to plugins when possible

sschr15:
This also makes a small change to [KotlinChooser](https://github.com/Vineflower/vineflower/blob/feature/semicolon-delegate/plugins/kotlin/src/main/java/org/vineflower/kotlin/KotlinChooser.java) to reduce the cost of `PluginContext.getLanguageSpec` calls by a significant amount


plugins/kotlin/src/main/java/org/vineflower/kotlin/KotlinChooser.java
```diff
@@ -43,7 +43,7 @@ public boolean isLanguage(StructClass cl) {
         StructAnnotationAttribute attr = cl.getAttribute((Key<StructAnnotationAttribute>) key);
         for (AnnotationExprent anno : attr.getAnnotations()) {
           if (anno.getClassName().equals("kotlin/Metadata")) {
-            setContextVariables(cl);
+//            setContextVariables(cl);
```
jaskarth, line 46: Was there a specific reason this is commented out? A comment here would be good.

jaskarth (approved):
Thanks a lot!

PR #472, opened 2025-03-23
Handle Kotlin's statements and their frequent sugar

sschr15:
This introduces large improvements to the structure of Kotlin decompilation. It gets rid of most (all?) incorrectly formatted labels, breaks, and continues, and it provides some support for `when` statements (those which compile to tableswitch and lookupswitch instructions).

This does make a couple small changes to the main Vineflower source.
It adds an additional getter in `SwitchHeadExprent`:
https://github.com/Vineflower/vineflower/blob/ab201e06b779b9c4f51127d24f182f5437678a89/src/org/jetbrains/java/decompiler/modules/decompiler/exps/SwitchHeadExprent.java#L145-L147
It also makes a visibility change in `ExprProcessor` (`private` to `public`):
https://github.com/Vineflower/vineflower/blob/ab201e06b779b9c4f51127d24f182f5437678a89/src/org/jetbrains/java/decompiler/modules/decompiler/ExprProcessor.java#L817

All other changes are Kotlin-specific.


jaskarth (approved):
Looks fine. Thanks a lot!

PR #473, opened 2025-03-23
Reduce filesizes, update Kotlin metadata to 2.1.20

sschr15:
Metadata jar generated with [the minifier](https://github.com/Vineflower/kt-metadata-minifier/tree/main).

This updates the Kotlin-related stuff to reference kotlin.metadata instead of kotlinx.metadata (upstream package change), and it updates the main build.gradle file to STORE plugin jars for better overall compression.


jaskarth (approved):
This looks great! Thanks a lot for this :)

PR #482, opened 2025-06-19
Improve decompiling fields

sschr15:
Cleans up field decompiling code for less repetition overall, and adds bytecode mappings to fields extracted from initializers


Kroppeb (approved):

jaskarth (approved):
Looks good! Thanks a lot for this work.

PR #484, opened 2025-07-02
 Compact canonical constructor support

Owen1212055:
[Minecraft diff](https://pastes.dev/KclRDfN7iR)

Would love feedback on the method pruning, not sure if there is a prefered way I should be doing this.
Additionally, I have to store a flag that a method wrapper is a compact constructor as I prune the fields making the check no longer valid later in the code.

### DIFF
--- a/src/org/jetbrains/java/decompiler/main/ClassWriter.java
+++ b/src/org/jetbrains/java/decompiler/main/ClassWriter.java
@@ -1543,7 +1543,8 @@ public class ClassWriter implements StatementWriter {
     boolean isRecord = cl.getRecordComponents() != null;
 
     // default constructor requires same accessibility flags. Exception: enum constructor which is always private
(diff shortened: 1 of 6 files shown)


src/org/jetbrains/java/decompiler/main/ClassWriter.java
```diff
@@ -1336,6 +1242,108 @@ else if (methodWrapper.varproc.getVarFinal(new VarVersionPair(index, 0)) == VarT
     return !hideMethod;
   }
 
+  private static int getParamCount(StructMethod mt, TextBuffer buffer, int indent, MethodWrapper methodWrapper, MethodDescriptor md, boolean isEnum, boolean init, boolean thisVar, GenericMethodDescriptor descriptor, int paramCount, boolean isInterface, int flags, StructClass cl) {
```
jaskarth, line 1245: I think this could use a slightly more descriptive name, since it has side effects as well as returning the param count.

PR #491, opened 2025-08-21
Remove public + static from class definitions in interfaces

Owen1212055:
Removes `public` and `static` from class definitions inside of interfaces.

MC diff:
https://pastes.dev/9egFieqcwL


jaskarth (approved):
Thanks a lot for the PR!

PR #493, opened 2025-09-04
Be more forgiving when parsing arguments for paths

### LINKED ISSUE #492: Specifying `--only` causes single-specified source to be treated as output
## Vineflower version
1.11.1

## Describe the bug
When invoking Vineflower with any of `--silent`, `--add-external`, or (as reported) `--only` and the only path provided is the source (that is, no output/destination is provided), Vineflower treats the path as an output. Thus, Vineflower does not see any sources provided and errors out with `error: no sources given`.

## Additional information

An example command-line invocation which triggers the bug:
```sh
java -jar vineflower-1.11.1.jar --only=org/jetbrains/java/decompiler/api/Decompiler vin
eflower-1.11.1.jar
```

The relevant code here is: https://github.com/Vineflower/vineflower/blob/cb7f62e9a85d390fcc97dbb475890cf1ad954fe9/src/org/jetbrains/java/decompiler/main/decompiler/ConsoleDecompiler.java#L127-L143

All of `--silent`, `--add-external` and `--only` ('pseudo-options' for the sake of reference) are treated by the same code that handles non-options, which also handles the sources and destination file paths. The logic is such that when more than two non-options are present, the last non-option is assumed to be an output destination.

However, when at least one pseudo-option is present alongside a source path at the end, the code then assumes the path is actually the output directory and skips it (corresponding to that `if (nonOptions > 1 ...)` check). Because of that, Vineflower technically sees no sources are specified, and proceeds to report that.

### Potential solution
The potential hotfix solution I see (though not having tested it) is to rework `nonOptions` to only be incremented and checked for file paths, being ignored for pseudo-options. That is, moving L127-L131 to within the `else` block.

However, if the last argument supplied is a pseudo-option, then this would interpret the last file path supplied to be a source rather than a output/destination.

A fix for that problem would involve holding onto the last file path while parsing, and registering it as a source or destination only if (a) another file path was encountered (thus, the held file path is a source) or (b) the end of parsing is reached (if there is already a source, then this is a destination; otherwise, it is the sole source).

_Thanks to Ami in the Vineflower spaces for originally reporting the issue._

sciwhiz12:
This PR fixes #492 by making the parsing of non-options (both file paths and the three pseudo-options `--only`, `--silent`, and `--add-external`) more forgiving when either (a) pseudo-options are provided and a single source is provided, or (b) pseudo-options are provided after the last path (source or destination).

This is an implementation of the potential solution I discussed in the linked issue's description.

I tested this with the following command-line invocations:
- `java -jar vineflower-1.12.0+local.jar --only=org/jetbrains/java/decompiler/api/Decompiler vineflower-1.11.1.jar` - Classes were outputted to console as expected.
- `java -jar vineflower-1.12.0+local.jar vineflower-1.11.1.jar --only=org/jetbrains/java/decompiler/api/Decompiler` - Classes were outputted to console as expected.
- `java -jar vineflower-1.12.0+local.jar vineflower-1.11.1.jar --only=org/jetbrains/java/decompiler/api/Decompiler out` - Classes were outputted to `out` as expected.
- `java -jar vineflower-1.12.0+local.jar vineflower-1.11.1.jar out --only=org/jetbrains/java/decompiler/api/Decompiler` - Classes were outputted to `out` as expected.

### DIFF
--- a/src/org/jetbrains/java/decompiler/main/decompiler/ConsoleDecompiler.java
+++ b/src/org/jetbrains/java/decompiler/main/decompiler/ConsoleDecompiler.java
@@ -90,7 +90,7 @@ public class ConsoleDecompiler implements /* IBytecodeProvider, */ IResultSaver,
 
     SaveType userSaveType = null;
     boolean isOption = true;
-    int nonOption = 0;
+    String lastPath = null;
     for (int i = 0; i < args.length; ++i) { // last parameter - destination
       String arg = args[i];
 
@@ -124,12 +124,6 @@ public class ConsoleDecompiler implements /* IBytecodeProvider, */ IResultSaver,
       }
 
       if (!parsed) {
-        nonOption++;
-        // Don't process this, as it is the output
-        if (nonOption > 1 && i == args.length - 1) {
-          break;
-        }
-
         isOption = false;
 
         if (arg.equals("-s") || arg.equals("--silent")) {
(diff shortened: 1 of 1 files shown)


jaskarth (approved):
Looks good to me, thanks a lot for the fix for this!

PR #494, opened 2025-09-08
Fix exception desync

Kroppeb:
Finally made a PR for this fix. AFAIK (almost) any input that would trigger this still fails to decompile.

The bug in question: when an exception handler is duplicated as a deobfuscation step, and the handler was inside another exception region, it did not have that fact properly copied over.


jaskarth (approved):
Thanks for the bugfix!

PR #497, opened 2025-10-21
Improve annotation formatting

Earthcomputer:
This PR improves annotation formatting by using the newline group system to decide when to split an annotation into multiple lines, rather than doing so unconditionally for name-value attribute pairs.

It also omits the `{}` braces for arrays in the decompiler output for annotation attributes, if the array is of length 1, and the single element of the array isn't itself an array.


sschr15 (approved):

jaskarth (approved):
Looks good, thanks a lot!

PR #502, opened 2025-10-27
Account members for enum + hide private for enum constructors

Owen1212055:


jaskarth (approved):
Looks good, thanks a lot for this!

PR #510, opened 2025-11-09
Aggressive memory optimizations

jaskarth:
Change sparse sets, sforms maps, and exprent bytecode offsets to prefer non-array backing implementations to prevent extremely sparse arrays and reduce peak memory usage when decompiling very large methods.


umbrageodotus:
Maybe make this optional?

aoqia194:
This PR is still targetting `develop/1.12.0` fyi!

PR #511, opened 2025-11-11
Fix decomp test failures caused by try/catch type changes

sschr15:
In Kotlin 2.4 and beyond, this may need revisiting because of the addition of [rich errors](https://resources.jetbrains.com/storage/products/kotlinconf-2025/may-22/Rich%20Errors%20in%20Kotlin%20_%20Michail%20Zarec%CC%8Censkij.pdf)


jaskarth (approved):
Looks good, thanks!

PR #515, opened 2025-11-15
Move metadata out of global state

sschr15:
Doing so also provides a strong base for future improvements to Kotlin decompilation


jaskarth (approved):
Thanks a lot for making this change!

PR #518, opened 2025-11-15
Test `lcmp`, `fcmp` and `dcmp` to `boolean` in TestNumberCompareToBoolean

Ampflower:
All the tests are functionally `number == number && number != 0`, baking in a NaN check.


jaskarth (approved):
Thanks for the tests!

PR #525, opened 2025-11-22
Kotlin: decompile most lambdas and anonymous objects

sschr15:
Decompiles many (most?) lambdas and anonymous objects partially as intended.

Variable captures do not decompile as expected currently. I probably have to figure out a renaming process to properly fix this.

This does break unhandled-by-metadata methods, but because Kotlin will (typically) publish metadata for everything meant to be shown to the end user, this is more of a side-effect on methods that should not be present in the first place.


jaskarth (approved):
Looks good, thank you for this work!

PR #530, opened 2025-12-17
Fix patterns with primitives and nested patterns in a switch

coehlrich:
Fixes patterns with primitives and nested patterns in a switch statement


jaskarth (approved):
Looks good, thanks!

PR #531, opened 2025-12-18
Fix primitives in pattern with an if statement with `false`

coehlrich:


jaskarth (commented):
Thanks for this! A unit test for this behavior would be good.

jaskarth (approved):

PR #532, opened 2025-12-19
Convert lambdas to method references when javac compiles a method reference to a lambda

coehlrich:
Converts lambdas to method references when javac compiles a method reference to a lambda, removes synthetic instance variable, and removes `Objects.requireNonNull`.

### DIFF
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
@@ -1,0 +1,266 @@
+package org.jetbrains.java.decompiler.modules.decompiler;
+
+import org.jetbrains.java.decompiler.code.CodeConstants;
+import org.jetbrains.java.decompiler.code.FullInstructionSequence;
+import org.jetbrains.java.decompiler.code.Instruction;
+import org.jetbrains.java.decompiler.main.ClassesProcessor;
+import org.jetbrains.java.decompiler.main.ClassesProcessor.ClassNode;
+import org.jetbrains.java.decompiler.main.ClassesProcessor.ClassNode.LambdaInformation;
+import org.jetbrains.java.decompiler.main.DecompilerContext;
+import org.jetbrains.java.decompiler.modules.decompiler.exps.*;
+import org.jetbrains.java.decompiler.modules.decompiler.stats.RootStatement;
+import org.jetbrains.java.decompiler.modules.decompiler.stats.Statement;
+import org.jetbrains.java.decompiler.struct.StructClass;
+import org.jetbrains.java.decompiler.struct.StructMethod;
+import org.jetbrains.java.decompiler.struct.consts.LinkConstant;
+import org.jetbrains.java.decompiler.util.InterpreterUtil;
+
+import java.io.IOException;
+import java.util.Iterator;
+
+public class MethodReferenceHelper {
+  public static boolean convertToMethodReference(RootStatement root) throws IOException {
+    return convertToMethodReferenceRec(root);
+  }
+
+  public static boolean convertToMethodReferenceRec(Statement stat) throws IOException {
+    boolean result = false;
+    for (Statement subStat : stat.getStats()) {
+      result |= convertToMethodReferenceRec(subStat);
+    }
+    if (stat.getExprents() != null) {
+      for (int i = 0; i < stat.getExprents().size(); i++) {
+        Exprent exp = stat.getExprents().get(i);
+        for (Exprent subExp : exp.getAllExprents(true)) {
+          if (convertToMethodReference(stat, i, subExp)) {
+            result |= true;
+          }
+          for (int j = 0; j < stat.getExprents().size(); j++) {
+            if (stat.getExprents().get(j) == exp) {
+              i = j;
+              break;
+            }
+          }
+          if (removeInstanceAssignment(stat, i, subExp)) {
+            result |= true;
+          }
+          for (int j = 0; j < stat.getExprents().size(); j++) {
+            if (stat.getExprents().get(j) == exp) {
+              i = j;
+              break;
+            }
+          }
+        }
+      }
+    }
+    return result;
+  }
+
+  private static boolean convertToMethodReference(Statement stat, int i, Exprent exp) throws IOException {
+    if (exp instanceof NewExprent newExprent
+        && newExprent.isLambda()
+        && !newExprent.isMethodReference()) {
+      if (stat.getTopParent().mt.getName().contains("<init>")) {
+        System.out.println();
+      }
+      String className = newExprent.getNewType().value;
+      ClassNode node = DecompilerContext.getClassProcessor().getMapRootClasses().get(className);
+      LambdaInformation info = node.lambdaInformation;
+      StructClass struct = DecompilerContext.getStructContext().getClass(info.content_class_name);
+      StructMethod method = struct.getMethod(info.content_method_key);
+      if (!method.hasModifier(CodeConstants.ACC_STATIC))
+        return false;
+      method.expandData(struct);
+      FullInstructionSequence seq = method.getInstructionSequence();
+      Iterator<Instruction> iterator = seq.iterator();
+      Instruction next = null;
+      int argument = 0;
+
+      if (!iterator.hasNext())
+        return false;
+      next = iterator.next();
+
+      // new instance
+      if (next.opcode == CodeConstants.opc_new) {
+        if (!iterator.hasNext())
+          return false;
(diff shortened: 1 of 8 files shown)


src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+    return result;
+  }
+
+  private static boolean convertToMethodReference(Statement stat, int i, Exprent exp) throws IOException {
+    if (exp instanceof NewExprent newExprent
+        && newExprent.isLambda()
+        && !newExprent.isMethodReference()) {
+      if (stat.getTopParent().mt.getName().contains("<init>")) {
+        System.out.println();
```
sschr15: Debug line

src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
```diff
@@ -0,0 +1,267 @@
...
+
+      // Load arguments
+      while (iterator.hasNext()
+          && next.opcode >= CodeConstants.opc_iload && next.opcode <= CodeConstants.opc_aload
+          && next.operand(0) == argument) {
+        next = iterator.next();
+        argument++;
+      }
+
+      if (next == null)
```
sschr15: This should never be true (`FullInstructionSequence` seems to only ever have non-null instructions). Did you mean `!iterator.hasNext()`?

src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
```diff
@@ -0,0 +1,267 @@
...
+            }
+          }
+        }
+      }
+    }
+    return result;
+  }
+
+  private static boolean convertToMethodReference(Statement stat, int i, Exprent exp) throws IOException {
+    if (exp instanceof NewExprent newExprent
```
sschr15: It's probably reasonable to flip this `if` to reduce a layer of nesting

src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
```diff
@@ -0,0 +1,267 @@
...
+import org.jetbrains.java.decompiler.struct.StructMethod;
+import org.jetbrains.java.decompiler.struct.consts.LinkConstant;
+import org.jetbrains.java.decompiler.util.InterpreterUtil;
+
+import java.io.IOException;
+import java.util.Iterator;
+
+public class MethodReferenceHelper {
+  public static boolean convertToMethodReference(RootStatement root) throws IOException {
+    return convertToMethodReferenceRec(root);
```
sschr15, line 23: This appears to also modify existing lambdas which could be represented with a method reference. Perhaps it should be behind a decompiler option

src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
```diff
@@ -0,0 +1,267 @@
...
+      ClassNode node = DecompilerContext.getClassProcessor().getMapRootClasses().get(className);
+      LambdaInformation info = node.lambdaInformation;
+      StructClass struct = DecompilerContext.getStructContext().getClass(info.content_class_name);
+      StructMethod method = struct.getMethod(info.content_method_key);
+      if (!method.hasModifier(CodeConstants.ACC_STATIC))
+        return false;
+      method.expandData(struct);
+      FullInstructionSequence seq = method.getInstructionSequence();
+      Iterator<Instruction> iterator = seq.iterator();
+      Instruction next = null;
```
sschr15: Unnecessary initializer

sschr15 (commented):
I have some small nitpicks, but overall it looks good to me.

~~manual bytecode parsing is scary~~

testData/results/pkg/TestGenericsQualified.dec
```diff
@@ -6,11 +6,11 @@ import java.util.concurrent.CompletableFuture;
 import java.util.stream.Stream;
 
 public class TestGenericsQualified {
-   public Comparator<String> field = Comparator.<String, Integer>comparing(s -> s.length()).thenComparing(i -> i.toString());// 10
+   public Comparator<String> field = Comparator.<String, Integer>comparing(s -> s.length()).thenComparing(String::toString);// 10
    public CompletableFuture<String> field2 = CompletableFuture.<String>supplyAsync(() -> "").thenCompose(s -> CompletableFuture.supplyAsync(() -> s + "2"));// 11
    public Optional<String> field3 = Optional.of("").map(s -> s + "3");// 12
-   public Stream<String> field4 = Stream.of("1", "2").sorted(Comparator.<String, Integer>comparing(s -> s.length()).thenComparing(i -> i.toString()));// 13 14
-   public Comparator<String> field5 = Comparator.comparing(String::length).thenComparing(i -> i.toString());// 15
+   public Stream<String> field4 = Stream.of("1", "2").sorted(Comparator.<String, Integer>comparing(s -> s.length()).thenComparing(String::toString));// 13 14
```
jaskarth, line 12: I believe this output is wrong. In the original source, these are regular lambdas and not references.

src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
```diff
@@ -0,0 +1,260 @@
...
+
+    // Load arguments
+    while (iterator.hasNext()
+        && next.opcode >= CodeConstants.opc_iload && next.opcode <= CodeConstants.opc_aload
+        && next.operand(0) == argument) {
+      next = iterator.next();
+      argument++;
+    }
+
+    boolean varargs = false;
```
jaskarth, line 101: What is this code doing?

src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
```diff
@@ -0,0 +1,260 @@
...
+        || newExprent.isMethodReference())
+      return false;
+    String className = newExprent.getNewType().value;
+    ClassNode node = DecompilerContext.getClassProcessor().getMapRootClasses().get(className);
+    LambdaInformation info = node.lambdaInformation;
+    StructClass struct = DecompilerContext.getStructContext().getClass(info.content_class_name);
+    StructMethod method = struct.getMethod(info.content_method_key);
+    if (!method.hasModifier(CodeConstants.ACC_STATIC))
+      return false;
+    method.expandData(struct);
```
jaskarth, line 71: Hmm, does this need to manually parse the instructions or would going through the exprents suffice?

jaskarth (changes requested):
Thanks a lot for working on this! Overall, I don't really understand the approach taken here. I'll look at the generated code and see if there might be a better way.

src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
coehlrich: The `ClassWrapper` on the `ClassNode` isn't set at this point and even if it was the exprents for lambdas are almost certainly not generated at this point since they are the last methods in the class file.

I did consider doing it when the lambda is written to the output like with array constructor references but that would prevent inlining the variable since if it happened when the lambda is written nothing in `codeToJava` would see that the variable assignment is removed whereas if the variable is removed during `codeToJava` then there I don't think there would be anyway for the variable to be inlined correctly since lambdas can't call a method on an instance without the instance being referenced via a variable.

src/org/jetbrains/java/decompiler/modules/decompiler/MethodReferenceHelper.java
coehlrich: The `varargs` variable is used for checking if the amount of method parameters of the lambda is what is expected.

testData/results/pkg/TestGenericsQualified.dec
coehlrich: They would have to match the lambda that javac generates when converting a method reference to a lambda.

I could require the variable to exist but then it would only work for java 25.

Toffikk:
Any updates on this PR? As the last comment is from over 4 months ago and this still doesn’t have any assigned priority and well, the changes made here are somewhat crucial for proper decompilation of bytecode emitted by java 25+

jaskarth:
The fix for the synthetic variable inlining has been pushed separately to develop/1.12. The fix for converting lambdas to method references needs further thought on coming up with a fix that doesn't impact non-reference lambdas.

PR #533, opened 2025-12-26
Add support for markdown javadoc

Earthcomputer:
Add support for Markdown Javadocs introduced in Java 23


jaskarth (approved):
Looks good. Eventually we'll want to move the Javadoc api into a more unified API, but for now this is fine.

PR #544, opened 2026-03-06
Simplify break structure for string switches

aoqia194:
This is cherry picked from my rewrite branch, so the diff is a little busted. You can view the proper diffs below.
Actual changes: https://github.com/aoqia194/vineflower/commit/cedf7038ae024272afb1df281a6fbb8fb9e6871d
Whitespace/style changes: https://github.com/aoqia194/vineflower/commit/5dbd2583954d182423d3067ed520bdcfdf836f39

It will remain a draft for a moment as I am still cleaning up some things such as TODOs that I have left myself, but I figured it would be faster to let people review it in advance, even once I resolve the TODOs.


aoqia194:
I've opened this for review as I've done everything I wanted to in relation to the TODOs. The code is still a bit messy and the whole of `integrateExits` will probably need a refactor after some time but that may just be out of the scope for this.

I also forgot to commit some other *.dec changes, which I will do right now.

src/org/jetbrains/java/decompiler/modules/decompiler/ExitHelper.java
```diff
@@ -217,6 +279,52 @@ private static Statement isExitEdge(StatEdge edge) {
     return null;
   }
 
+  private static Statement isRealIfEdge(IfStatement ifst, StatEdge edge) {
```
jaskarth: A comment here on what "real" means, as well as what the function returns would be good.

src/org/jetbrains/java/decompiler/modules/decompiler/ExitHelper.java
```diff
@@ -217,6 +279,52 @@ private static Statement isExitEdge(StatEdge edge) {
...
+    Statement dest = edge.getDestination();
+
+    // NOTE(aoqia): A cheap check to avoid processing unwanted things (see TestSwitchLoop#test with this check disabled)
```
jaskarth:
```suggestion
    // A cheap check to avoid processing unwanted things (see TestSwitchLoop#test with this check disabled)
```

src/org/jetbrains/java/decompiler/modules/decompiler/ExitHelper.java
```diff
@@ -106,87 +100,151 @@ private static int integrateExits(Statement stat) {
...
+                    ifst.removeSuccessor(realIfedge);
+                  }
+
+                  ifst.addSuccessor(new StatEdge(StatEdge.TYPE_BREAK, ifst, defaultCase, ifedge.closure));
+                }
+              }
+            }
+          }
+
+          StatEdge newedge = new StatEdge(StatEdge.TYPE_REGULAR, ifst.getFirst(), bstat);
```
jaskarth: A description of this block of code would be good.

jaskarth (commented):
Thanks a lot for these changes! This looks good, I've made a note of a couple of documentation changes that would be good.

src/org/jetbrains/java/decompiler/modules/decompiler/ExitHelper.java
aoqia194: I tried to add some more clarifying comments. Let me know if it's sufficient. Though I myself don't quite exactly know what the code block for L173+ does, because I didn't write it, but I can assume it's just adding the newly-created statement and things.

src/org/jetbrains/java/decompiler/modules/decompiler/ExitHelper.java
aoqia194: I couldn't exactly think of a really clear name for it, but I've called it the "if-break edge" and you can see the explanation with the code snippet in the docstring.

aoqia194:
I didn't mean to re-request the review, sorry about that. It was meant to be a reminder to resolve the issue if you think it's fixed lol.

jaskarth (approved):
This is great, thanks a lot for the update!

PR #545, opened 2026-03-06
Add helper case functions

aoqia194:
Very simple helper functions for case manipulation for switches. Will change a little later when I add a case class to SwitchStatement to further clean it up.

(cherry picked from commit decdb6efbe849567c4dbb01bf82a653cf39e2b80)

### DIFF
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/stats/SwitchStatement.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/stats/SwitchStatement.java
@@ -584,7 +584,7 @@ public class SwitchStatement extends Statement {
     this.getCaseStatements().add(index, stat);
     this.getCaseValues().add(index, !isAddDefault ? values : Collections.singletonList(null));
 
-    if (values != null) {
+    if (!isAddDefault) {
       var newCaseEdgeLst = new ArrayList<StatEdge>();
       for (int i = 0; i < values.size(); i++) {
         final var edge = new StatEdge(StatEdge.TYPE_REGULAR, this.getBasichead(), stat);
@@ -592,10 +592,10 @@ public class SwitchStatement extends Statement {
 
         this.getBasichead().addSuccessor(edge);
       }
-      this.getCaseEdges().add(index, newCaseEdgeLst);
-    }
 
-    if (isAddDefault) {
+      this.getCaseEdges().add(index, newCaseEdgeLst);
+    } else {
+      // Just add the default and move on!
       final var edge = new StatEdge(StatEdge.TYPE_REGULAR, this.getBasichead(), stat);
       this.setDefaultEdge(edge);
     }
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/stats/SwitchStatement.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/stats/SwitchStatement.java
@@ -1,6 +1,8 @@
 // Copyright 2000-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
 package org.jetbrains.java.decompiler.modules.decompiler.stats;
 
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.jetbrains.java.decompiler.code.SwitchInstruction;
 import org.jetbrains.java.decompiler.code.cfg.BasicBlock;
 import org.jetbrains.java.decompiler.main.DecompilerContext;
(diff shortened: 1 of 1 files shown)


src/org/jetbrains/java/decompiler/modules/decompiler/stats/SwitchStatement.java
```diff
@@ -535,4 +536,80 @@ public void scopeCaseStatement(Statement stat) {
...
+    final var index = this.getCaseStatements().size();
+    this.addCaseInternal(index, null, stat);
+  }
+
+  public void addCase(List<@NotNull Exprent> values, Statement stat) {
+    final var caseCount = this.getCaseStatements().size();
+    this.addCaseInternal(this.getDefaultEdge() != null ? caseCount - 1 : caseCount, values, stat);
+  }
+
+  public void addCase(int index, List<@NotNull Exprent> values, Statement stat) {
```
jaskarth: Hmm, should this be `List<@NotNull Exprent>` or `@Nullable List<Exprent> values`? Since in `addCaseInternal` you're checking if the list itself can be null, it should be the latter for both this and the `addCase` above.

jaskarth (commented):
Thanks for the additional helper methods! Just one comment from me.

src/org/jetbrains/java/decompiler/modules/decompiler/stats/SwitchStatement.java
aoqia194: My thought process for this (which is probably a bad way of going about it in hindsight) is:
- A null `values` list means that `addCaseInternal` will add a default case (as seen in `setDefaultCase`).
- A non-null `values` list with null values implies a default case (since a default case will always have the case list of `List<@Nullable Exprent>` or always a `List<null>` aka a List with only 1 value inside of it, that being null. This can't be the case for `addCase` functions because they specifically do NOT add default cases (which should only be done using `setDefaultCase` and will add one if needed).
- A non-null `values` list with non-null values implies just a normal case list, eg having a case value list of `ConstExprent`s, this should be the input for `addCase`.

aoqia194: I should add some comments to explain this

jaskarth: I see, I think that means the definition in `addCaseInternal` should be `@Nullable List<@Nullable Exprent> values`, since it's a list that can be null that contains exprents which can be null. Some docs for that method would also be good.

aoqia194: I actually got the explanation to you mixed up as well, this goes to show how unacquainted I am with these annotations. It can be either a null `values` list or a non-null `values` list with non-null values. So I think `@Nullable List<Exprent> values` is good, or `@Nullable List<@NotNull Exprent>` of which I don't know which is better.

jaskarth: Ah, usually tools these days assume not-null as the default so `@Nullable List<Exprent> values` should be fine.

aoqia194:
Ok I've done all the changes I wanted to do, this should be the final state unless more feedback is given.

jaskarth:
Thanks for making the changes! The last style change that's needed here is that we generally avoid `var` and use the explicit type, and don't use `final` for local variables/parameters.

aoqia194:
> Thanks for making the changes! The last style change that's needed here is that we generally avoid `var` and use the explicit type, and don't use `final` for local variables/parameters.

I was curious about how var is used in the project. I used it a lot, so I'll be sure to change it everywhere else in my other branches too.

jaskarth:
There are uses of it in some parts of the code as a lot of different developers/projects are a part of the codebase. My general view is that `var` reduces readability of the code especially when not using an IDE, such as when looking at code at GitHub. `final` for local variables and parameters specifically is just syntax sugar, and in my opinion hurts readability.

jaskarth (approved):
Looks good, thanks a lot!

aoqia194:
> There are uses of it in some parts of the code as a lot of different developers/projects are a part of the codebase. My general view is that `var` reduces readability of the code especially when not using an IDE, such as when looking at code at GitHub. `final` for local variables and parameters specifically is just syntax sugar, and in my opinion hurts readability.

I agree generally. There are some cases though where it has merit but yeah. I am also used to final all my vars but only because sometimes I've accidentally overwritten local vars that shouldn't be changed, but I do understand what you mean. Anyhow, thanks for merging.

PR #546, opened 2026-03-07
Using pattern matching

Kroppeb:
So I cleaned up `SimplifyExprentsHelper` a bit. Mostly converting `instanceof` checks followed by casts into pattern matching.
This resulted in being able to combine nested if statements in a bunch of places.

Now, part of me wants to do a full reformat of the file, but on the other hand, given how we don't have custom record deconstuctors yet, I ended up manually indenting the large if blocks to make them clearer.

A lot of the changes are going to be indentation changes so it might be advised to ignore whitespace when reviewing this.

There shouldn't be any logic changes. I did change to order of some comparisons in some places, removed null checks followed by instanceof checkes and used pattern variables holding the value already instead of calling the getter again.


jaskarth (commented):
Thanks for the cleanup! This is a minor code style thing, but I would prefer if statements' opening and closing parentheses on the same line as the condition.

jaskarth (approved):
Looks good, thanks!

PR #549, opened 2026-03-24
Fix infinite looping during generics checking

sschr15:
This may also improve generics checks when lambdas are involved


jaskarth (approved):
Great, thanks a lot!

PR #550, opened 2026-03-24
Handle object construction and related constructs

sschr15:
**Modifies [`NewExprent`](https://github.com/Vineflower/vineflower/blob/77353d83c3501b88b9831f0368794fbda8f0311e/src/org/jetbrains/java/decompiler/modules/decompiler/exps/NewExprent.java)**: to publicize a method

Most visibly removes `new` from constructor calls, but also converts arrays to actually being `arrayOf` and similar


jaskarth (approved):
Looks good, thanks for these fixes!

PR #551, opened 2026-03-25
Fix incorrectly renamed variables

coehlrich:
There are some compile errors where some variable reads in catches are using the renamed variables when they are not renamed


jaskarth (approved):
Thanks a lot for this, looks good!

PR #552, opened 2026-03-25
Kotlin metadata rework, part 2

sschr15:
The Big One <sup>excluding all the other big ones</sup>

Moves Kotlin metadata onto the items that they represent, instead of single large blobs of everything on the `StructClass` objects. 
This is beneficial for further improvements, notably library metadata isolation and resugaring method calls


jaskarth (approved):
Looks good, thanks a lot for this work!

PR #554, opened 2026-03-26
fix finally processing in plugins

Kroppeb:


sschr15 (approved):
test results do look cleaner than before, but apparently i also broke some things at some point

jaskarth (approved):
I can't believe this oversight, thanks a lot for fixing this.

PR #555, opened 2026-03-27
Fix incorrectly renaming pattern match variables

coehlrich:


src/org/jetbrains/java/decompiler/modules/decompiler/vars/VarDefinitionHelper.java
```diff
@@ -1958,10 +1976,61 @@ private void iterateClashingExprent(Statement stat, StructMethod mt, Map<Stateme
         }
       }
     }
+    
+    if (exprent instanceof FunctionExprent function) {
+      iterate = false;
+      // https://docs.oracle.com/javase/specs/jls/se25/html/jls-6.html#jls-6.3.1
+      switch (function.getFuncType()) {
+      case BOOLEAN_AND -> {
```
jaskarth: In the VF codestyle, case statements are indented by 1.

jaskarth (commented):
Thanks for this work! The patch itself looks fine, but I would like to see some new tests to exercise the logic that's added for pattern matches, ternaries, and loops.

jaskarth (approved):
Thanks for the update!

PR #558, opened 2026-03-30
Rewrite string-switch processing

### LINKED ISSUE #500: Switch-on-string fails to simplify
## Vineflower version
master

## Describe the bug
Given a simple `return switch` with the condition being a string and the switch block only containing one case and the required default, it will not properly simplify like a switch-on-string with two cases correctly simplifies. The same thing happens when given a switch that has string cases and returns something.

## Additional information
Can be easily replicated following the decompilation output of TestSingleCaseStrSwitch, TestEclipseSwitchString and TestSwitchExprString1.

aoqia194:
Rewrites and refactors string-switch processing in SwitchHelper to handle a lot more edge cases easier. Also adds a simple test class that tests some common switch patterns which can also be found in other tests too, bundling them was just more ideal.

Fixes #500!

#### Some personal notes

- The result records that I created are there because I needed to return multiple values from a method and I didn't know what the clearest way to do it was, and using a record worked well for my situation (though it isn't the prettiest).

- I did get overly ambitious with making the `StringSwitch` interface overly generic. My reason in the moment was that I wasn't sure if it was necessary to expand the processing even further in the future to cover more cases or not, and I still really wanted clear separate processing for individual string-switch 'types'.

- I tried to refactor reused code into separate methods as best as I could!

As with anything, I'm 100% OK with changing anything in here, so please leave as many review comments as you want.


src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
```diff
@@ -199,144 +193,449 @@ private static boolean simplify(SwitchStatement switchStatement, StructMethod mt
...
+   * @param switchHeadValue the value of the switch head exprent
+   * @return true if simplified successfully, false otherwise
+   */
+  private static boolean trySimplifyStringSwitch(SwitchStatement switchStat, Exprent switchHeadValue) {
+    // Get the type of switch by matching against each type
+    Optional<StringSwitch> result = StringSwitch.match(switchStat);
+    if (result.isEmpty()) {
+      return false;
+    }
+    StringSwitch switchInfo = result.get();
```
aoqia194: This optional here might be in a hot path, but im not sure

sschr15:
Hi, thanks for working with string switches! ~~Unfortunately, Kotlin string `when` blocks do not decompile as expected in this regard. I've pushed some tests for it (also enabling the switch code, which @jaskarth and I apparently missed when writing the Kotlin plugin). If possible, I'd love to see these tests get decompiled as expected as well~~ (we've discussed to push this off to a future PR, so all is good as-is)

aoqia194:
This latest commit was just because I went over the code from #545 and realised I forgot to also add the default edge to the switch's case edges list. I also removed an unnecessary thing I did with nullable list.

src/org/jetbrains/java/decompiler/modules/decompiler/stats/SwitchStatement.java
```diff
@@ -553,7 +553,7 @@ public void setDefaultCase(Statement stat) {
       return;
     }
 
-    this.addCaseInternal(0, null, stat);
+    this.addCaseInternal(0, Collections.singletonList(null), stat);
```
jaskarth, line 556:
```suggestion
    this.addCaseInternal(0, List.of(null), stat);
```

src/org/jetbrains/java/decompiler/modules/decompiler/stats/SwitchStatement.java
```diff
@@ -576,15 +576,18 @@ public void addCase(int index, List<Exprent> values, Statement stat) {
...
+  private void addCaseInternal(int index, List<@Nullable Exprent> values, Statement stat) {
     // Basichead is always the first stat
     this.getStats().add(1 + index, stat);
     this.getCaseStatements().add(index, stat);
-    this.getCaseValues().add(index, !isAddDefault ? values : Collections.singletonList(null));
+    this.getCaseValues().add(index, values);
 
-    if (!isAddDefault) {
+    // If it's a default case, or a regular case
+    if (values.size() == 1 && values.get(0) == null) {
```
jaskarth, line 586: Hmm, what happens to the logic here if null appears in a list of size greater than 1? I see that the attached commit mentioned it fixes a bug, it'd be good to write a unit test for it.

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
```diff
@@ -598,119 +867,238 @@ public static <T extends Exprent> void findExprents(Statement start, Class<? ext
...
-              }
-            }
-          }
-        }
-      }
+  private sealed interface StringSwitch permits Split, InlineSplit, NullableSplit, Merged {
+    SwitchStatement first();
+    SwitchStatement target();
+
+    static Optional<StringSwitch> match(SwitchStatement stat) {
```
jaskarth: I think this method should be rewritten to not use optional and lambdas. It obfuscates the logic a little bit, and calling lambdas that capture local variables allocates an object every time.

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
```diff
@@ -199,144 +193,428 @@ private static boolean simplify(SwitchStatement switchStatement, StructMethod mt
...
+      // Remove all case data for this case as we create new values/edges/stats later.
+      switchInfo.first().getCaseEdges().get(processedCasesIdx).forEach(StatEdge::remove);
+      switchInfo.first().getCaseEdges().remove(processedCasesIdx);
+      switchInfo.first().getCaseValues().remove(processedCasesIdx);
+      switchInfo.first().getCaseStatements().remove(processedCasesIdx);
+      // First stat is always basichead
+      switchInfo.first().getStats().remove(currIf);
+
+      // Remove normal edge between case and default block if it exists
+      {
```
jaskarth: Hm, is there any reason for a block here?

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
```diff
@@ -199,144 +193,428 @@ private static boolean simplify(SwitchStatement switchStatement, StructMethod mt
...
+
+    return true;
+  }
+
+  private static boolean unwrapMergedSwitchCases(StringSwitch switchInfo) {
+    int processedCasesIdx = 0;
+
+    List<Statement> caseStats = switchInfo.first().getCaseStatements();
+    int origCaseCount = caseStats.size();
+    // noinspection ForLoopReplaceableByForEach
```
jaskarth:
```suggestion
```

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
```diff
@@ -199,144 +193,428 @@ private static boolean simplify(SwitchStatement switchStatement, StructMethod mt
...
+    SyntheticDupVarResult result = _result.get();
+
+    // Replace synthetic var in switch head
+    result.switchHead().replaceExprent(result.switchHead().getValue(), result.realVar());
+    if (!(switchInfo instanceof NullableSplit)) {
+      result.headExprs().remove(result.headExprs().size() - result.dupVarIdx());
+    }
+  }
+
+  // A temporary structure to store result from getSyntheticDupVar, but it's too annoying to figure out a better way.
```
jaskarth:
```suggestion
```

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
```diff
@@ -199,144 +193,428 @@ private static boolean simplify(SwitchStatement switchStatement, StructMethod mt
...
+        }
+
+        if (condition instanceof InvocationExprent condInvoc && condInvoc.getLstParameters().size() == 1) {
+          List<Exprent> ifExprs = ifStat.getIfstat().getExprents();
+          Exprent ifEqFirstExpr = ifExprs.get(0);
+          Exprent realVal = condInvoc.getLstParameters().get(0);
+
+          // Split switches are guaranteed to have a unique intermediary even with hashcode collision.
+          // Merged switches, however, may contain multiple case labels/strings sharing 1 intermediary.
+          //   This is commonly found where switches have explicit return values.
```
jaskarth:
```suggestion
          // This is commonly found where switches have explicit return values.
```

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
```diff
@@ -199,144 +193,428 @@ private static boolean simplify(SwitchStatement switchStatement, StructMethod mt
...
+        Exprent targetVal = targetCaseVal.get(j);
+        StatEdge targetEdge = targetCaseEdges.get(j);
+
+        Integer intermediate = targetVal instanceof ConstExprent e ? e.getIntValue() : null;
+
+        // If the real return val exists (in the case of merged switches usually) then create new cases for them.
+        List<Exprent> realVal = caseMap.get(intermediate);
+        // If the real case value is null, it means this case wasn't linked between
+        //   both first and second switches and is thus an invalid case (can be safely removed).
+        // As noted from Jasmine (adf75bb):
```
jaskarth: I think it'd be better to mention the place where the comment is rather than the commit, and to remove the name.

src/org/jetbrains/java/decompiler/util/collections/VBStyleCollection.java
```diff
@@ -29,8 +29,13 @@ public boolean add(E element) {
   }
 
   @Override
-  public boolean remove(Object element) {   // TODO: error on void remove(E element)
-    throw new RuntimeException("not implemented!");
+  public boolean remove(Object element) {
```
jaskarth: I think this should be reverted. VBStyleCollection is soft-deprecated and `removeWithKey` should have the equivalent functionality, so I think it'd be better to avoid changing it.

jaskarth (changes requested):
Thanks a lot for doing this refactoring! I've listed some feedback here, most of it is related to documentation and style.

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
aoqia194: This was a side effect from previous code, I did have some conflicting variables previously so the block was to scope the variables. I've since refactored it so I can now remove that block, good catch.

src/org/jetbrains/java/decompiler/modules/decompiler/stats/SwitchStatement.java
aoqia194: Are you sure this is OK? IDEA warns me about passing `null` when it is annotated with `@NotNull`, and also the documentation of `of` states:
"Throws: `NullPointerException` – if an element is `null` or if the array is `null`"

jaskarth: Nevermind, I forgot about that restriction :)

src/org/jetbrains/java/decompiler/modules/decompiler/SwitchHelper.java
aoqia194: Do you want say the line number from that commit? I don't understand how I can reference the place the comment originally was when that place wont exist with the comment after this PR

jaskarth: Ah, in that case just remove the reference since it looks like you've reproduced the commit text below.

src/org/jetbrains/java/decompiler/modules/decompiler/stats/SwitchStatement.java
aoqia194: Resolved, not necessary?

aoqia194:
I've resolved pretty much everything I can, but GitHub is still saying there are requested changes. I can't figure it out so I've just requested another review 🫠

jaskarth (approved):
Thanks for the work!

PR #559, opened 2026-04-08
Add Automatic-Module-Name to the manifest

### LINKED ISSUE #121: Define Automatic-Module-Name in Manifest
It would be good practice to define a stable Automatic-Module-Name for projects that want to depend on QF as a module.
Even though Quilt doesn't use it anymore Matcher is such a project and it even has a comment regarding this: https://github.com/QuiltMC/matcher/blob/main/build.gradle#L60.

What should the module name be?
- `org.jetbrains.java.decompiler`
- `org.quiltmc.quiltflower`
- Something else?

TropheusJ:
Gives VF an automatic module name of `org.jetbrains.java.decompiler`, so it can be used in modular projects without gymnastics.

I've manually tested that:
- The jar is still not modular when run standalone from CLI
- The module is named correctly at runtime in a modular context
- The plugins built into the all-jar are still able to be loaded in a modular context

Resolves #121

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -189,7 +189,8 @@ jar {
     attributes (
       'Main-Class': 'org.jetbrains.java.decompiler.main.decompiler.ConsoleDecompiler',
       'Implementation-Name': "Vineflower",
-      'Implementation-Version': project.version
+      'Implementation-Version': project.version,
+      'Automatic-Module-Name': "org.vineflower.vineflower"
     )
   }
 }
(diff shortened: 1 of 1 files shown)


sschr15 (approved):
Makes sense to me

zml2008:
I think it would make more sense to match the module name that upstream FF uses, to let VF be a bit closer to a drop-in replacement

TropheusJ:
forgot they had one, done

jaskarth (approved):
Looks good, thank you!

PR #561, opened 2026-04-09
Fix switch expressions with a return not having their while true loop removed

coehlrich:


jaskarth (approved):
Looks fine, thanks for the fix!

PR #563, opened 2026-04-14
Fix switch expressions not being inlined when none of the case statements directly contain `yield`

coehlrich:


jaskarth (approved):
Looks good!

PR #564, opened 2026-04-15
Fix not inlining cases with multiple switch expressions in a single line

coehlrich:
Fixes cases like
```java
    public int test(int i1, int i2) {
        return switch (i1) {
            case 0 -> 1;
            case 1 -> 0;
            default -> 0;
        } + switch (i2) {
            case 0 -> 1;
            case 1 -> 0;
            default -> 0;
        };
    }
```
decompiling to
```java
    public int test(int i1, int i2) {
        byte var10000 = switch (i1) {
            case 0 -> 1;
            case 1 -> 0;
            default -> 0;
        };

        return var10000 + switch (i2) {
            case 0 -> 1;
            case 1 -> 0;
            default -> 0;
        };
    }
```


utafrali (commented):
The core fix is logically sound: refactoring traversal into `iterateStatementsNested` correctly scopes phantom switch processing, and the two `SwitchExprent` special-cases in the child-exprent iterators are the right place to stop inlining at switch expression boundaries. The main actionable issues are a missing cycle guard in the inner while loop (real risk with adversarial bytecode) and an unused `IOException` import in the new test source file.

testData/src/java16/pkg/TestSwitchExpressionMultiple.java
```diff
@@ -0,0 +1,23 @@
+package pkg;
+
+public class TestSwitchExpressionMultiple {
```
utafrali, line 3: `import java.io.IOException;` is unused — nothing in this file throws or references `IOException`. Remove it.

src/org/jetbrains/java/decompiler/modules/decompiler/StackVarsProcessor.java
```diff
@@ -156,8 +180,51 @@ private static boolean iterateStatements(RootStatement root, SSAUConstructorSpar
...
+              nodes = newNodes;
+            } while (changed);
+
+            if (nodes.size() == 1) {
+              ndsucc = nodes.get(0);
+            } else {
+              ndsucc = null;
+            }
+
+            while (ndsucc != null && ndsucc.exprents.isEmpty()) {
```
utafrali: This `while` loop has no cycle detection. If a sequence of empty-exprents nodes forms a cycle in the graph (degenerate but possible with malformed bytecode), this will spin forever. Add a visited set here, similar to the `setVisited` used in the outer loop:

```java
Set<DirectNode> emptyVisited = new HashSet<>();
while (ndsucc != null && ndsucc.exprents.isEmpty() && emptyVisited.add(ndsucc)) {
    List<DirectEdge> next = ndsucc.getSuccessors(DirectEdgeType.REGULAR);
    if (next.size() == 1) {
        ndsucc = next.get(0).getDestination();
    } else {
        ndsucc = null;
    }
}
```

src/org/jetbrains/java/decompiler/modules/decompiler/StackVarsProcessor.java
```diff
@@ -123,21 +121,47 @@ private static boolean iterateStatements(RootStatement root, SSAUConstructorSpar
...
+        }
+      }
+
+      return 0;
+    });
+
+    return res;
+  }
+
+  private static boolean iterateStatementsNested(Statement root, DirectNode start, SSAUConstructorSparseEx ssa, StackSimplifyOptions options, DirectGraph dgraph, Map<VarVersionPair, LocalVariable> lvts, Set<DirectNode> setVisited, LinkedList<DirectNode> parentStack, LinkedList<Map<VarVersionPair, Exprent>> parentStackMaps, Map<VarVersionPair, Exprent> currentParentStackMap) {
```
utafrali: The `iterateStatementsNested` signature now takes 10 parameters, several of which are mutable output channels (`parentStack`, `parentStackMaps`). This makes call sites hard to read and the contract hard to reason about. A small inner class or record holding `{setVisited, parentStack, parentStackMaps, currentParentStackMap}` would clarify what is shared state vs. what is per-call input. Not blocking for this fix, but worth a follow-up.

src/org/jetbrains/java/decompiler/modules/decompiler/StackVarsProcessor.java
```diff
@@ -156,8 +180,51 @@ private static boolean iterateStatements(RootStatement root, SSAUConstructorSpar
         DirectNode ndsucc = succs.get(0).getDestination();
 
         if (ndsucc.type == DirectNodeType.TAIL && !ndsucc.exprents.isEmpty()) {
-          lstLists.add(succs.get(0).getDestination().exprents);
-          nd = ndsucc;
+
+          // Skip over phantom switch statements
+          if (ndsucc.statement instanceof SwitchStatement switchSt && switchSt.isPhantom()) {
```
utafrali: The outer guard `!ndsucc.exprents.isEmpty()` prevents entering the phantom-switch traversal block when the TAIL node has no exprents. If a phantom switch statement's TAIL node happens to be empty (e.g., an intermediate structural node), the traversal is skipped entirely and the multiple-switch-expression case would still not be inlined. Consider whether the phantom-switch detection should be separated from the `!ndsucc.exprents.isEmpty()` guard, or add a comment explaining why an empty TAIL can never be a phantom switch in practice.

src/org/jetbrains/java/decompiler/modules/decompiler/StackVarsProcessor.java
```diff
@@ -123,21 +121,47 @@ private static boolean iterateStatements(RootStatement root, SSAUConstructorSpar
...
+      
+      // Treat phantom switch statements as a sub method
+      if (nd.statement instanceof SwitchStatement switchSt && switchSt.isPhantom() && root != switchSt) {
+        res |= iterateStatementsNested(switchSt, nd, ssa, options, dgraph, lvts, setVisited, stack, stackMaps, mapVarValues);
+        continue;
+      } else if (!root.containsStatement(nd.statement) && !(root instanceof RootStatement)) {
+        parentStack.add(nd);
+        parentStackMaps.add(new HashMap<>(currentParentStackMap));
+        continue;
+      }
```
Kroppeb: I don't understand this case

src/org/jetbrains/java/decompiler/modules/decompiler/StackVarsProcessor.java
Kroppeb: I don't understand what is going on here :[

PR #567, opened 2026-04-29
Fix a bunch of variables not getting their LVTs

coehlrich:
Changes `processBlock` to use the offsets stored in `Instruction`s instead of the old offsets


jaskarth (approved):
Comparing a few of the offsets, this looks like a solid improvement. Thanks for this fix!

PR #570, opened 2026-05-06
Reuse existing logic to fix wrong ppmm inline match

### LINKED ISSUE #569: Side-effecting decrement gets moved into a `&&` short-circuit
## Vineflower version
Latest upstream

## Describe the bug
In this case the issue is already tested (although with the wrong conclusion) in Vineflower itself, see how https://github.com/Vineflower/vineflower/blob/master/testData/src/java8/pkg/TestIfLoop.java#L10 has
```java
    public int testCompoundCondition(int i, int j) {
        while (i > 0 && j < 3) {
            if (j < 0) {
                j -= 1;

                if (i > 3) {
                    if (j == -2) {
                        j = 1;
                    }
                }
            }

            if (i > 10) {
                i--;
                return i;
            }
        }

        return 3;
    }
```

But the decompile in https://github.com/Vineflower/vineflower/blob/master/testData/results/pkg/TestIfLoop.dec#L10 has:
```java
public int testCompoundCondition(int i, int j) {
      while (i > 0 && j < 3) {
         if (j < 0) {
            if (i > 3 && --j == -2) {
               j = 1;
            }
         }

         if (i > 10) {
            return i - 1;
         }
      }

      return 3;
   }

```

These two are semantically different, because the original would eventually return 3 due to `int` wraparound, but the decompiled version would run forever, for example with:
```
new pkg.TestIfLoop().testCompoundCondition(1, Integer.MIN_VALUE)
```

I hit the issue while decompiling one of the old J2ME-era `.jar`'s, so I can't paste its classes here, but the test is already self-evident.

## Additional information
I'm developing a fork of Vineflower focused on J2ME decompilation with LLM agents, and my goal is to ensure that the decompiled code is 100% semantically equivalent to the original bytecode. Vineflower has been a great help because it already covers 98% of all of the work, the rest is almost looking good. I have read the Vineflower's AI policy so I won't PR any of the LLM-written fixes, but could I still report issues if they are clearly in the Vineflower core, not specifically obscure J2Me bytecode?

Kroppeb:
Replaces the custom logic in `inlinePPIandMMIIf` by one already existing in ``. That version got a patch for better handling of `&&`, `||` and ternaries. 

The `inlinePPIandMMIIf` was also moved as it opens up optimizations for `IfHelper`

Fixes #569


jaskarth (approved):
Looks good, thank you for the cleanup!

Issue #568, opened 2026-04-29
Parsing Failure on nested Try-Finally statement


heisluft:
## Vineflower version
1.12.0

## Describe the bug
Vineflower fails to decompile a method with supposedly nested try blocks

```
java.lang.RuntimeException: parsing failure!
  at org.jetbrains.java.decompiler.modules.decompiler.decompose.DomHelper.parseGraph(DomHelper.java:211)
  at org.jetbrains.java.decompiler.main.rels.MethodProcessor.codeToJava(MethodProcessor.java:174)
```

## Additional information
FF also fails on this.
Procyon gives this output:
```java
     public final void run() {
        try {
            final ArrayList list = new ArrayList();
            final URL url = new URL("350e7af8f93185cd68da36455f205091b307bf20");
            final BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(url.openStream()));
            String line;
            while ((line = bufferedReader.readLine()) != null) {
                list.add(line);
            }
            bufferedReader.close();
            for (int i = 0; i < list.size(); ++i) {
                final URL url2 = url;
                final String s = (String)list.get(i);
                final URL context = url2;
                Label_0334: {
                    try {
                        final String[] split;
                        final String child = (split = s.split("626b0ce09264a7f53caa9b0c2bf6ccffdee7048a"))[0];
                        final int int1 = Integer.parseInt(split[1]);
                        Long.parseLong(split[2]);
                        final File file;
                        if (!(file = new File(this.a, child)).exists() || file.length() != int1) {
                            file.getParentFile().mkdirs();
                            this.a(new URL(context, child.replaceAll("bf9e96420f44eaf37bedec73110b00011e48b385", "49b6a324c23dee23e4b4d2ec32c07a58d5a865f0")), file);
                            if (this.c) {
                                break Label_0334;
                            }
                        }
                        final d b = this.b;
                        final String s2 = child;
                        final File file2 = file;
                        final String s3 = s2;
                        final d d = b;
                        final int index = s3.indexOf("1600bb70e4ca936f3e61ecdc7a2916dfb965d238");
                        final String substring = s3.substring(0, index);
                        final String substring2 = s3.substring(index + 1);
                        if (substring.equalsIgnoreCase("a041afffb1915ac44607a64bee8752c2b72ca743")) {
                            d.x.a(substring2, file2);
                        }
                        else if (substring.equalsIgnoreCase("952165729b27207181d32c1e5114ed7be714131c")) {
[...]

PR #585, opened 2026-05-14
Add debug marker exceptions

Kroppeb:
These are 2 exceptions that are interpreted differently.

* `CatchAllException`: These get decompiled as exception type `null` instead, just like the catch created for `finally`
* `UnCatchException`: Removes all catch handlers on a statement outside. This is used to create a test case for #568 without needing jasm.


src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
```diff
@@ -418,6 +418,11 @@ public interface IFernflowerPreferences {
   @Type(DecompilerOption.Type.STRING)
   String METHOD_TO_DECOMPILE = "method-to-decompile";
 
+  @Name("DEBUG: Enable special handling of marker exceptions")
```
jaskarth, line 421: Since this is mainly for unit testing, we shouldn't expose it to the public. That means we shouldn't have the Name, Description, and Type annotations here (see `UNIT_TEST_MODE`)

src/org/jetbrains/java/decompiler/struct/StructMethod.java
```diff
@@ -343,11 +344,26 @@ else if (opcode >= opc_invokevirtual && opcode <= opc_invokestatic) {
         exceptionClass = null;
       } else {
         exceptionClass = pool.getPrimitiveConstant(excClass).getString();
+        if (DecompilerContext.getOption(DEBUG_MARKER_EXCEPTIONS)) {
+          if (exceptionClass.equals("vineflower/markerexception/CatchAllException")){
```
jaskarth:
```suggestion
          if ("vineflower/markerexception/CatchAllException".equals(exceptionClass)){
```
Just for consistency :)

testData/src/java8/vineflower/markerexception/CatchAllException.java
```diff
@@ -0,0 +1,7 @@
+package vineflower.markerexception;
```
jaskarth: Hmm, I don't think this is the right place to put these classes since it means we can only use it in the java 8 source set. Can this go into the test fixture source set? Also, I'd prefer it if the package was `org.vineflower.marker`.

jaskarth (changes requested):
Looks neat! I've added some comments here.

src/org/jetbrains/java/decompiler/main/extern/IFernflowerPreferences.java
Kroppeb: Ah, didn't realize that was an option. I'll do that.

testData/src/java8/vineflower/markerexception/CatchAllException.java
Kroppeb: The test data classes don't have a dependency on test fixtures. I just tries and that doesn't work, as test fixtures is compiled with Java 17. 

I can add a small extra source set that holds these 2 classes instead?

PR #588, opened 2026-05-20
FinallyProcessor: guard instrOldOffsets removals when out of sync with seq

FewzY:
## Describe the bug

`FinallyProcessor.removeExceptionInstructionsEx` mutates two parallel structures on the same `BasicBlock`: `InstructionSequence seq` and `List<Integer> instrOldOffsets`. The two can drift apart when `verifyFinallyEx` re-invokes this method on the same handler block during multi-pass iteration — a previous pass shrank one list more than the other, then a later pass tries to remove from the now-empty offsets list and throws.

### Minimal repro

A Kotlin `bitmap.compress(...).use { ... }` try-finally (the standard \`CloseableKt.closeFinally\` pattern) consistently triggers it on R8-shrunk Android code.

### Stack trace

```
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
  at java.base/java.util.ArrayList.remove(ArrayList.java:552)
  at org.jetbrains.java.decompiler.modules.decompiler.FinallyProcessor.removeExceptionInstructionsEx(FinallyProcessor.java:1058)
  at org.jetbrains.java.decompiler.modules.decompiler.FinallyProcessor.verifyFinallyEx(FinallyProcessor.java:573)
  at org.jetbrains.java.decompiler.modules.decompiler.FinallyProcessor.iterateGraph(FinallyProcessor.java:90)
```

## Describe the fix

`InstructionSequence.removeLast()` already self-guards (only removes if non-empty). I extended the same pattern to the four `instrOldOffsets.remove(...)` call sites and the two `seq.removeInstruction(0)` sites inside `removeExceptionInstructionsEx`. No behaviour change when the lists are in sync — only graceful no-op when they aren't.

This is a band-aid, not a root-cause fix: the underlying invariant violation (the two lists falling out of sync across re-invocations) is still worth investigating. But shipping the guard prevents loss of the entire method body for code that the rest of the pipeline could otherwise decompile correctly.

## Provenance disclosure

## Checklist
- [ ] Unit tests were added or modified to validate this change. *(Existing test suite passes; happy to add a fixture if you'd like.)*

### DIFF
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/FinallyProcessor.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/FinallyProcessor.java
@@ -1049,25 +1049,35 @@ public class FinallyProcessor {
     if (finallytype == 3) { // empty finally handler
       for (int i = seq.length() - 1; i >= 0; i--) {
         seq.removeInstruction(i);
-        instrOldOffsets.remove(i);
+        if (i < instrOldOffsets.size()) {
+          instrOldOffsets.remove(i);
+        }
       }
     } else {
       if ((blocktype & 1) > 0) { // first
         if (finallytype == 2 || finallytype == 1) { // astore or pop
-          seq.removeInstruction(0);
-          instrOldOffsets.remove(0);
+          if (!seq.isEmpty()) {
+            seq.removeInstruction(0);
+          }
+          if (!instrOldOffsets.isEmpty()) {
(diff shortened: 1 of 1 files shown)


Kroppeb:
Could you attach a class file that shows the behavior. I don't have the tools to recreate your example.

jaskarth:

Issue #470, opened 2025-03-10
Decompilation failure with kotlin `crossinline` lambdas crossinlined into a suspending context


solonovamax:
## Vineflower version
1.11.1

## Describe the bug
Vineflower cannot decompile kotlin `crossinline` lambdas when they are passed to a function which takes a suspending lambda.

## Additional information
Here is some code which reproduces the issue. I have minimized it as much as possible, and have not been successful in minimizing it any further.
```kotlin
fun test() {
    return test2 { }
}

inline fun test2(crossinline fn: () -> Unit) {
    return runBlocking { fn() }
}
```

Here is the stack trace:
```
ERROR: Class ca/solostudios/nyx/plugin/compile/TestKt$test$$inlined$test2$1 couldn't be fully decompiled.
java.lang.IllegalStateException: Couldn't find method <anonymous> (Lkotlinx/coroutines/CoroutineScope;)V in class ca/solostudios/nyx/plugin/compile/TestKt$test$$inlined$test2$1
	at org.vineflower.kotlin.struct.KFunction.parse(KFunction.java:112)
	at org.vineflower.kotlin.KotlinWriter.writeClass(KotlinWriter.java:221)
	at org.jetbrains.java.decompiler.main.ClassesProcessor.writeClass(ClassesProcessor.java:500)
	at org.jetbrains.java.decompiler.main.Fernflower.getClassContent(Fernflower.java:196)
	at org.jetbrains.java.decompiler.struct.ContextUnit.lambda$save$3(ContextUnit.java:195)
	at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(ForkJoinTask.java:1403)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)
```

If you run `javap` on the .class file, this is what it emits:
```
[...]

Issue #262, opened 2023-04-16
Big Kotlin Work List


sschr15:
A bunch of things to get Kotlin to be (mostly) decompiling Kotlin and not a Kotlin-Java hybrid, split into a few main categories.

# Technical Backend
- [ ] Publish Kotlin metadata to Kotlin-targeting code
- [ ] Kotlin exprent equivalents
- [ ] Other special Kotlin construct handling code

# Non-method fixes
- [ ] Field declarations
- [ ] `data class` support
- [ ] `annotation class` support
- [ ] Primary constructors
- [ ] Other constructors (constructors currently decompile as normal functions)
- [ ] `extends` / `implements` clauses
- [ ] Anonymous classes
- [ ] Annotation arrays should use `[]`
- [ ] Other explicit arrays should use `arrayOf` or the equivalent primitive type
- [ ] All known-to-have-generic types must specify generics
- [ ] Higher level functions not based on SAM interfaces
- [ ] Top-level declarations

# Method fixes
- [ ] Remove unnecessary semicolons (which should be all of them if decompiling reasonably)
- [ ] Remove `new` from constructor calls
- [ ] Destructuring declarations
- [ ] String interpolation
- [ ] Smart Casts
- [ ] Receiver types (should decompile as `this` or as `this@...`)
- [ ] SAM lambda decompilation
- [ ] Non-SAM lambda decompilation 
- [ ] `for` loops and `if` conditions on ranges
  - [ ] `repeat`
  - [ ] `rangeTo`
  - [ ] `until` and later Kotlin `rangeUntil`
  - [ ] `downTo`
  - [ ] `step`
- [ ] Infix functions
- [ ] Operator functions
- [ ] `tailrec` functions
- [ ] `when` statements

Kroppeb:
`lateinit var`

LlamaLad7:
Inline function call decompilation would be cool. It should be possible from the source map.

sschr15:
> Inline function call decompilation would be cool. It should be possible from the source map.

SMAP is not capable enough to indicate individual names of inlined functions, just the files and locations within those files for source information. Kotlin local variables do tend to include local variable entries that could be helpful, but I need to look more into it

PR #589, opened 2026-05-20
Kotlin: fall back to JVM supertypes for anonymous classes without Kotlin metadata

FewzY:
## Describe the bug

When `KotlinWriter` encounters a `ClassNode.Type.ANONYMOUS` whose `KElement` attribute is not a `KClass`, it throws `IllegalStateException(\"Anonymous class does not have Class Kotlin metadata\")` and the surrounding class's body is lost.

This is the dominant Kotlin-plugin failure on batch-mode decompilation of R8-shrunk Android apps. On a real ~3700-class app jar (Dartsmind 4.4.1 + 4.3.13), this single throw produced **302 distinct failing classes**, all from patterns the Kotlin team has already documented in #470 and the broader Kotlin work list at #262:

- Anonymous comparators from `sortedWith(compareByDescending { … })`
- Inlined suspending lambdas (\`crossinline\` into a suspending context — see #470's stack-trace migration from \`KFunction.parse:112\` (1.11.1) to \`KotlinWriter.writeClassDefinition:742\` (1.12.0))

The user's reported bisection finding for #470 — *\"the bug only triggers in batch mode; passing the outer class alone with the jar via `-e=...` works\"* — is consistent with this throw: in single-class mode, the lambda gets a `KFunction` attribute earlier in the call chain (line 292: `writeLambda(node, ...)`) so it never reaches `writeClassDefinition`. In batch mode, ordering or state differences leave some lambdas without that attribute, and the throw aborts them.

## Describe the fix

Rather than fix the upstream `KFunction`/`KClass` attribute attachment (which I don't have the full picture on without more time in `KotlinChooser.parseMetadataFor` and `KFunction.parse`), this PR makes `writeClassDefinition` graceful when no `KClass` metadata is present:

- The `KClass` branch is unchanged.
- The fallback branch emits `object : <super>, <iface1>, <iface2>` using the JVM `StructClass`'s superclass and interface list — the same information the Java-side writer uses for anonymous classes. `Object`/`Any` is filtered out as before.

### Measured impact on a real Android app

| metric | before | after |
| --- | --- | --- |
| Files with `\$VF: Couldn't be decompiled` (whole jar) | 593 | 455 |
| Of those, `IllegalStateException` from line 742 | 302 | 0 |
| Files in app's own package (com/zaglab) with any failure | 47 / 958 | 1 / 958 |

The remaining 455 jar-wide failures are all unrelated Java-core bugs (parsing failures, `FlattenStatementsHelper` NPEs, `IfHelper.collapseIfIf` NPE, etc.) — none in the Kotlin plugin's hot path.

### Caveats

- This is graceful **fallback**, not a metadata-attachment fix. The underlying \"why doesn't this class have `KClass` metadata?\" question remains open and is worth investigating separately — at minimum, the path in `KotlinChooser.parseMetadataFor` for k=3 synthetic classes where `KFunction.parse` `continue`s on unmatched suspend `<anonymous>` (line 113-117 of `KFunction.java`) leaves the class without an attribute and seems like a likely culprit.
- The fallback output won't recover Kotlin-style reified generics or other metadata-only info, but it does produce *something compilable* with the right supertype shape, which is a huge win over losing the entire enclosing class body.

## Related issues

- #470 — same root error site, also unresolved
- #262 — \"Big Kotlin Work List\" master tracker

## Provenance disclosure

I diagnosed and authored this patch while pair-coding with Claude (Anthropic's coding agent) inside a focused diagnose-loop session against a real-world Android APK. I have personally read, understood, validated, and signed off on the change. Flagging this because of the AI-tools checkbox on the bugfix template — happy to discuss further, rewrite/refactor in your preferred style, or close this and re-file as an issue with minimal repros if you'd rather not accept AI-assisted PRs at all.

## Checklist
- [ ] Unit tests were added or modified to validate this change. *(Existing Kotlin test suite passes including \`TestAnonymousEverything\`. Happy to write a new fixture that covers the metadata-absent path if you'd like.)*

### DIFF
--- a/plugins/kotlin/src/main/java/org/vineflower/kotlin/KotlinWriter.java
+++ b/plugins/kotlin/src/main/java/org/vineflower/kotlin/KotlinWriter.java
@@ -738,28 +738,48 @@ public class KotlinWriter implements StatementWriter, Flags {
 
   private void writeClassDefinition(ClassNode node, TextBuffer buffer, int indent, KElement ktData, int kotlinFlags) {
     if (node.type == ClassNode.Type.ANONYMOUS) {
-      if (!(ktData instanceof KClass cls)) {
-        throw new IllegalStateException("Anonymous class does not have Class Kotlin metadata");
-      }
-
       buffer.append("object");
 
-      boolean first = true;
-      for (ProtoBuf.Type supertype : cls.proto().getSupertypeList()) {
-        KType kType = KType.from(supertype, cls.resolver());
-        if (VarType.VARTYPE_OBJECT.equals(kType)) {
-          // skip Any / Object supertype
-          continue;
-        }
+      if (ktData instanceof KClass cls) {
+        boolean first = true;
+        for (ProtoBuf.Type supertype : cls.proto().getSupertypeList()) {
+          KType kType = KType.from(supertype, cls.resolver());
+          if (VarType.VARTYPE_OBJECT.equals(kType)) {
+            // skip Any / Object supertype
+            continue;
+          }
 
-        if (first) {
-          buffer.append(" : ");
-          first = false;
-        } else {
-          buffer.append(", ");
-        }
+          if (first) {
+            buffer.append(" : ");
+            first = false;
+          } else {
+            buffer.append(", ");
+          }
 
-        buffer.append(kType.stringify(indent));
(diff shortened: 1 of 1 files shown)


jaskarth:

PR #594, opened 2026-05-22
Fix ClassesProcessor not whitelisting class files

### LINKED ISSUE #593: Failed to save class when decompiling specific class files in a JAR
## Vineflower version
1.12.0

## Describe the bug
When decompiling a JAR and using the `--only=my/class` parameter, vineflower will fail to output the decompiled java files with an exception being raised "Failed to save class"

## Additional information

Stacktrace:
```java
Exception in thread "main" java.lang.RuntimeException: Failed to save class
	at org.jetbrains.java.decompiler.main.decompiler.DirectoryResultSaver.saveClassEntry(DirectoryResultSaver.java:31)
	at org.jetbrains.java.decompiler.main.extern.IResultSaver.saveClassEntry(IResultSaver.java:25)
	at org.jetbrains.java.decompiler.struct.JarContextSource$1.acceptClass(JarContextSource.java:128)
	at org.jetbrains.java.decompiler.struct.ContextUnit.save(ContextUnit.java:236)
	at org.jetbrains.java.decompiler.struct.StructContext.saveContext(StructContext.java:178)
	at org.jetbrains.java.decompiler.main.Fernflower.decompileContext(Fernflower.java:155)
	at org.jetbrains.java.decompiler.main.decompiler.ConsoleDecompiler.decompileContext(ConsoleDecompiler.java:250)
	at org.jetbrains.java.decompiler.main.decompiler.ConsoleDecompiler.main(ConsoleDecompiler.java:195)
Caused by: java.nio.file.NoSuchFileException: /home/aoqia/Projects/git/local/leaf/out/920353448238085945/decompiled/zombie/GitVersion.java
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:108)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:113)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newFileChannel(UnixFileSystemProvider.java:214)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:245)
	at java.base/java.nio.file.spi.FileSystemProvider.newOutputStream(FileSystemProvider.java:426)
	at java.base/java.nio.file.Files.newOutputStream(Files.java:215)
	at java.base/java.nio.file.Files.newBufferedWriter(Files.java:2729)
	at java.base/java.nio.file.Files.newBufferedWriter(Files.java:2770)
	at org.jetbrains.java.decompiler.main.decompiler.DirectoryResultSaver.saveClassEntry(DirectoryResultSaver.java:26)
	... 7 more
```

aoqia194:
Fixes #593!

This PR branch should probably be tested on the CLI against some other projects, because I am not sure if adding this second check will influence anything.

### An explanation about the bug

When using `--only=com/example`, the `com/example` is marked as a whitelisted prefix; thus any directories in a JAR being matched against these whitelist prefixes will be created. However, when matching a singular class file (e.g. `--only=com/example/MyClass`), the original check against the to-be-created directory `com/example` would fail.

The check will boil down to:
```java
"com/example".startsWith("com/example/MyClass")
```
and it is obvious that this would fail, thus the directory would not be created, and after decompilation happens and the class file is being dumped to a file, it will fail because of the parent directories not existing.

It took me a few hours to find a fix that didn't feel hacky, though I will admit that this still feels a little hacky and might have some unknown-to-me consequences. This simple change will affect this whitelist check and makes the check:
```java
"com/example".startsWith("com/example/MainClass") || "com/example/MainClass".startsWith("com/example")
```
This check satisfies the directory-creating code and will create the directories just fine. Yippee!

### DIFF
--- a/src/org/jetbrains/java/decompiler/main/ClassesProcessor.java
+++ b/src/org/jetbrains/java/decompiler/main/ClassesProcessor.java
@@ -82,7 +82,7 @@ public class ClassesProcessor implements CodeConstants {
       return true;
 
     for (String prefix : this.whitelist) {
-      if (cls.startsWith(prefix))
+      if (cls.startsWith(prefix) || prefix.startsWith(cls))
         return true;
     }
 


aoqia194:
I forgot to mention that it's also possible to just check the folders exist again right before writing the file, but I think this might be cheaper anyway? Also it may be better internally that these folders are marked as whitelisted in the first place.

aoqia194:
I'm converting this to draft because this same issue has come up again on this branch despite this fixing it previously?

aoqia194:
Ok it seems that this still works after checking, I think it was just user error with my vineflower wrapper. Maybe to prevent false-positives in the whitelisting with this extra check, we can also check for a suffix like `.java` or `.class` to determine whether a single file is being whitelisted or not? I don't think it's necessary right now however.

PR #595, opened 2026-05-31
Fix non-ambiguous varargs overload being marked as ambiguous

### LINKED ISSUE #584: Non ambiguous varargs overload are not simplified sometimes
## Vineflower version
a303602c822276e9ce55dfe9f06b04aef545e66b

## Describe the bug
```Java
private void testVarargs() {
    this.add("first id");
    this.add(new String[]{"second id"}); // ambiguous

    this.add(5); // not ambiguous yet vineflower consider it ambiguous because of the two other overloads
    // will decompile to the longer version: add(new Integer[]{5});
}

public void add(String id) {
}

public void add(String... ids) {
}

public void add(Integer... ids) {
}
```
decompiles to
```Java
private void testVarargs() {
    this.add("first id");
    this.add(new String[]{"second id"});
    this.add(new Integer[]{5});
}

public void add(String id) {
}

public void add(String... ids) {
}

public void add(Integer... ids) {
}
```

The last add call is not simplified to this.add(5) unlike before, it seems Vineflower consider it ambiguous because of the two other overloads.
I believe this is a small regression since https://github.com/Vineflower/vineflower/commit/bc57845546f2db7b3b6aa3533d798e12f21c711e which fixes some ambiguous overload not being correctly called.

## Additional information
Compiled using openjdk 25.0.3

aoqia194:
Fixes #584!

I'm not sure if this is the right way to go about it since I know nothing about the area that this code lives. From what I've seen, the issue is that when it is checking a vararg type is ambiguous, it will match even when the type left type and right type are Object.

The main point is to check if the type can be boxed implicitly, but this probably isn't the cleanest way (if you know better, please comment or suggest!).

I've added a test case to show this, but I'll explain it here. In the example shown in the issue, there are overloads for `String`, `String...`, and `Integer...`. When the code is checking for ambiguous vararg overload types, and given `overload(123);`, it will incorrectly match both `String...` and `Integer...` as types because they are both `Object`; thus it becomes ambiguous even though it isn't exactly.

Uhh thats really all of it, simple change. All i did was when checking that the types are matching, I also checked the intersect/meet point between the types. if they don't share a common type anywhere at all then it's probably some completely different type and it shouldn't even be considered!

### DIFF
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/exps/InvocationExprent.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/exps/InvocationExprent.java
@@ -1395,7 +1395,7 @@ public class InvocationExprent extends Exprent {
           }
 
           // If we have two types which are both Object, it's hard to tell what is compatible.
-          // We need to get the common type between the two. If there isn't, it means the type isn't compatible.
+          // We need to get the common type between the two. If there isn't one, it means the type isn't compatible.
           if (leftType.type == CodeType.OBJECT && rightType.type == CodeType.OBJECT) {
             VarType commonType = VarType.meet(leftType, rightType);
             if (commonType == null || commonType == VarType.VARTYPE_NULL) {
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/exps/InvocationExprent.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/exps/InvocationExprent.java
@@ -1393,6 +1393,15 @@ public class InvocationExprent extends Exprent {
           if (leftType.arrayDim != rightType.arrayDim) {
             return false;
           }
+
+          // If we have two types which are both Object, it's hard to tell what is compatibe.
+          // We need to get the common type between the two. If there isn't, it means the type isn't compatible.
+          if (leftType.type == CodeType.OBJECT && rightType.type == CodeType.OBJECT) {
+            VarType commonType = VarType.meet(leftType, rightType);
+            if (commonType == null || commonType == VarType.VARTYPE_NULL) {
+              return false;
+            }
+          }
         }
       }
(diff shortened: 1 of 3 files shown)


testData/results/pkg/TestArrayArg.dec
aoqia194, line 1: this diff is terrible!

src/org/jetbrains/java/decompiler/modules/decompiler/exps/InvocationExprent.java
```diff
@@ -1393,6 +1393,15 @@ private boolean matchesVarargs(VarType[] left, VarType[] right, StructMethod lef
           if (leftType.arrayDim != rightType.arrayDim) {
             return false;
           }
+
+          // If we have two types which are both Object, it's hard to tell what is compatibe.
```
jaskarth:
```suggestion
          // If we have two types which are both Object, it's hard to tell what is compatible.
```

src/org/jetbrains/java/decompiler/modules/decompiler/exps/InvocationExprent.java
```diff
@@ -1393,6 +1393,15 @@ private boolean matchesVarargs(VarType[] left, VarType[] right, StructMethod lef
...
+          // We need to get the common type between the two. If there isn't, it means the type isn't compatible.
```
jaskarth:
```suggestion
          // We need to get the common type between the two. If there isn't one, it means the type isn't compatible.
```

jaskarth (approved):
Looks good! Just two typos from me:

jaskarth (approved):
Looks good! I think this logic should still work in cases where the decompiler can't find one or both classes, which was my main concern. Thanks for the contribution!

PR #599, opened 2026-06-08
Remove finally semaphores.

Kroppeb:
I don't like the semaphores, they produce confusing code without any reason, sometimes resulting in code that does not compile.


jaskarth:
I'm not sure if this is the right approach. The semaphores serve to show that there is an expected try/finally structure, but Vineflower was unable to resugar it due to bugs in the finally processor.

PR #601, opened 2026-06-22
Add support for new Eclipse string-switch variant

### LINKED ISSUE #133: Eclipse switch-on-string support
Eclipse produces different switch on string, as seen in `TestEclipseSwitchString`. We should resugar those properly as well as string switch produced by javac.

aoqia194:
Fixes #133

I've done some minor refactoring in this as well, so I apologise in advance. Most of those are just variable renames or deleting redundant comments.

Also fixes some minor bugs that caused other string-switches to fail the beautification! (what is this called lmao)


testData/results/TestEclipseSwitchString2.dec
```diff
@@ -6,21 +6,13 @@ public class TestEclipseSwitchString2 {
...
-            if (var2.equals("3")) {
-               var1 = "d";// 27
-            }
+      switch (var2 = TestEclipseSwitchString2.Inner.get()) {// 19
+         case "1":
+            var1 = "b";// 21
+         case "2":
+            var1 = "c";// 24
+         case "3":
+            var1 = "d";// 27
```
aoqia194, line 17: I didn't notice this at first, but this switch doesn't have breaks in the case list. Would this mean that it's not correct because of fallthrough?

aoqia194:
I am opening this back up for review as I've finally found and fixed the mistakes I've done! (thanks Kropp for the help with the desync and validation issues)

jaskarth (approved):
Looks good to me, thank you for the PR!

PR #605, opened 2026-07-04
Fix merged variable keeping its narrower declared type

### LINKED ISSUE #578: Byte switch variable reused as int causes invalid Java
## Vineflower version
1.12.0

## Describe the bug
Original Java:
```java
public class Repro {
  static byte[] bytes = new byte[4];

  public static int test(int[] box) {
    int value;
    if (box[0] < bytes.length) {
      value = bytes[box[0]];
    } else {
      value = 15;
    }

    switch ((byte)value) {
      case 2:
        value = box[0];
        return value + 6;
      default:
        return 0;
    }
  }
}
```

Decompiled:
```java
public class Repro {
   static byte[] bytes = new byte[4];

   public static int test(int[] var0) {
      byte var1;
      if (var0[0] < bytes.length) {
         var1 = bytes[var0[0]];
      } else {
         var1 = 15;
      }

      switch ((byte)var1) {
         case 2:
            var1 = var0[0];
            return var1 + 6;
         default:
            return 0;
      }
   }
}
```

This fails to compile with:
```
error: incompatible types: possible lossy conversion from int to byte
            var1 = var0[0];
```

## Additional information
Java source compiled with `OpenJDK 64-Bit Server VM Temurin-25.0.3+9 (build 25.0.3+9-LTS, mixed mode, sharing)`, `--release 8`.

lorenzo-cingano:
Fixes #578.

A variable that is a byte on one path and reused as an int on another was declared as a byte, so the decompiled code did not compile. Assigning an int to a byte needs a narrowing cast (JLS 5.1.3).

When two variable versions are merged, `remapVar` retypes the writes and the type map to the wider type, but it only refreshed the declared type for declarations kept in a statement's var definitions list. When the merge target's declaration is stored inline as a definition exprent, it kept its old, narrower type. This adds the missing refresh for that case.

### Testing
Added `TestByteVarIntMerge` (java8nodebug), which declares the variable as a byte before the fix and an int after.

This also updates the expected output of `TestSwitchExpressionAssignmentOrder`. Its committed result already had the same problem baked in (`byte var3` assigned an int, `byte var4` assigned int arithmetic) and did not compile; both now decompile as int. A separate assignment order issue in that test's `test4` is pre existing and left untouched.

The full test suite passes.

### DIFF
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/vars/VarDefinitionHelper.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/vars/VarDefinitionHelper.java
@@ -1041,6 +1041,15 @@ public class VarDefinitionHelper {
       else if (expr instanceof VarExprent) {
         VarExprent var = (VarExprent)expr;
         VarVersionPair old = new VarVersionPair(var);
+        if (old.equals(to) && var.isDefinition()) {
+          // Refresh the merge target's declared type when its declaration is stored inline,
+          // so it matches the widened type of the merged writes (Vineflower#578).
+          VarType merged = getMergedType(from, to);
+          if (merged != null) {
+            var.setVarType(merged);
+          }
+          continue;
+        }
         if (!old.equals(from)) {
           continue;
         }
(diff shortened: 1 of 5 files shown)


jaskarth (approved):
This fix looks good, thank you for the contribution!

PR #609, opened 2026-07-17
Fix incorrect compact record constructors

### LINKED ISSUE #608: Canonical constructor with field assignment using ternary does not appear in output
## Vineflower version
1.12.0

## Describe the bug
A record coded and compiled with the following constraints does not properly decompile with its canonical constructor:
- Compiled with `javac -g` (Generate all debugging info).
- Its canonical constructor is fully specified with the formal parameter list.
- There are at least two record components.
- Each field is assigned with its corresponding parameter in the same order as the parameter list.
- At least one field is assigned with a ternary operation (`?:`).

The compilation method, the fully specificed canonical constructor, the number of components being greater than 2, and the order of assignments matter

The condition and the results of the ternary operation do not matter. Whether there is more than one field assigned with a ternary does not matter. Names do not matter.

Compiled on `javac 25.0.3`.

## Additional information
Source code:
```java
public record Repro(long a, long b) {
    public Repro(long a, long b) {
        this.a = ("a".equals("b")) ? 0 : 1;
        this.b = b;
    }
}
```

Vineflower output:
```java
==== Repro.java ====
public record Repro(long a, long b) {
}
```

coehlrich:
Fixes #608


jaskarth (approved):
Looks good, thanks!

PR #612, opened 2026-08-05
Fix switches with both types and enum values

coehlrich:
Fixes switches with both types and enum values

### DIFF
--- a/src/org/jetbrains/java/decompiler/modules/decompiler/exps/FieldExprent.java
+++ b/src/org/jetbrains/java/decompiler/modules/decompiler/exps/FieldExprent.java
@@ -157,10 +157,6 @@ public class FieldExprent extends Exprent {
   public TextBuffer toJava(int indent) {
     TextBuffer buf = new TextBuffer();
 
-    if (wasCondy) {
-      buf.append("/* $VF: constant dynamic */ ");
-    }
-
     if (isStatic) {
       if (name.equals("TYPE") && ExprUtil.PRIMITIVE_TYPES.containsKey(classname)) {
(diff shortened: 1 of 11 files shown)


electronicboy:
This fixes the decompilation entirely failing with NoiseChunk, but the output is missing the class label and so isn't re-compilable

src/org/jetbrains/java/decompiler/modules/decompiler/exps/FieldExprent.java
```diff
@@ -157,10 +157,6 @@ private boolean isAmbiguous() {
   public TextBuffer toJava(int indent) {
     TextBuffer buf = new TextBuffer();
 
-    if (wasCondy) {
```
jaskarth, line 160: It would be good to keep this. If this shows up in switch specific code, could a flag be set to suppress it when it's resugared?

testData/src/java25/pkg/TestSwitchPatternMatchingJ25.java
```diff
@@ -64,4 +64,17 @@ public String test6(Object o) {
       default -> throw new RuntimeException();
     };
   }
+
+  enum Test7 {
+    VALUE1,
+    VALUE2;
+  }
+  public String test7(Object o) {
```
jaskarth, line 72: It'd be good to add a test that uses a when clause, to make sure that those work with this too.

jaskarth (commented):
Thanks a lot for this PR! The logic looks good, I just have two comments here.

jaskarth (approved):
Looks good, thanks again for this PR!

